PR #3, opened 2020-01-21
Setup models

juliasaveliff:
- Moved existing `model.py` into new `djreact/models` directory 
- Created initial models with fields for Instance, Model, Rack, User


juliasaveliff:
@benauriemma

PR #4, opened 2020-01-21
migrated model changes to DB and updated readme

juliasaveliff:
`rackcity/migrations/0002_itinstance_itmodel_rack_user.py` is generated automatically with migration to postgres DB


juliasaveliff:
@clairecfu

PR #12, opened 2020-01-24
added an endpoint models/add that accepts a post request, performs basic validation, and saves a new model

benauriemma:


benauriemma:
CircleCI is failing but not sure why

rackcity/views/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .it_model_views import model_detail, model_list
+from .it_model_views import model_detail, model_list, model_add
```
benauriemma: Is it possible to do something like:
```
from .it_model_views import *
```

rackcity/views/it_model_views.py
```diff
@@ -21,6 +21,21 @@ def model_list(request):
...
     #     return JsonResponse(serializer.errors, status=400)
 
+def model_add(request): # TODO: make it only accept POSTs
+    """
+    Add a new model
+    """
+    if request.method == 'POST':
+        data = JSONParser().parse(request)
+        serializer = ITModelSerializer(data=data)
+        if serializer.is_valid(): # TODO: figure out what .valid() does.  Different fields are required for add, update, etc
```
benauriemma: There's more logic that should be done here.  For example, checking that the vendor+model_num is unique.  It should also allow you to leave model_id blank and fill it in itself (maybe that should even be required?)

benauriemma: Maybe each endpoint should have a function that encapsulates its validation and returns usable errors when validation fails?

benauriemma:
This isn't meant to be the final form of the endpoint! Just a POC and to use for testing

benauriemma:
OH ALSO please try this out on Postman before you approve. I had to do some funky stuff with Cookies so if it's broken for you I'll write up the process before merging

rackcity/views/__init__.py
juliasaveliff: technically yes, but I think it's general best practice to avoid wildcard imports (even if we are importing everything from that module anyway)

rackcity/views/it_model_views.py
juliasaveliff: Good idea. Also we can enforce the `model_id` field to be unique on the Django model (as an argument on the field), but I don't think the front end should have to pass that in with this POST, since it's always going to be concatenated, so I think we would want to do that in this method?

juliasaveliff: On second thought...I don't think we need `model_id` at all since it's sort of redundant. We can use this on `vendor` and `model_number`

https://docs.djangoproject.com/en/dev/ref/models/options/#unique-together

clairecfu:
circleci tests were failing because the config files weren't in your branch, but now they are!

PR #15, opened 2020-01-24
added auth (it was like 70% done) and it is now required for the model_list endpoint

benauriemma:


juliasaveliff:
@clairecfu slacked you a relevant notion page to explain what to do with these changes! ready 2 be merged.

PR #23, opened 2020-01-26
fixed no data bug

clairecfu:
Before, if no data was being returned on models/racks/instances, front end was failing to load anything on the landing page. Should now display "no data found" and a spinner icon. This was only discovered when we deployed on staging because staging db has no data in it.


juliasaveliff:
cool

PR #25, opened 2020-01-26
Final changes to models

juliasaveliff:
Updated models to reflect changes here

https://www.notion.so/Database-Tables-Proposal-89116f694f3f45bfb3f1d7ff5d14444f

Also added notional data


create_notional_data.txt
```diff
@@ -0,0 +1,105 @@
+from rackcity.models import ITInstance, ITModel, Rack
+Run `python manage.py shell` and run the following commands
```
juliasaveliff: Couldn't figure out how to make this a script that would run with the right settings, so keeping it as .txt file for now

rackcity/admin.py
```diff
@@ -1,5 +1,7 @@
 from django.contrib import admin
 
-from .models import Article
+from .models import ITInstance, ITModel, Rack
 
-admin.site.register(Article)
+admin.site.register(ITInstance)
+admin.site.register(ITModel)
+admin.site.register(Rack)
```
juliasaveliff: can see these from Django `/admin` endpoint on localhost

rackcity/api/serializers.py
```diff
@@ -1,51 +1,35 @@
...
-    class Meta:
-        model = Article
-        fields = ('id', 'title', 'content')
-
-
 class ITInstanceSerializer(serializers.ModelSerializer):
     class Meta:
         model = ITInstance
-        fields = ('hostname', 'height', 'model', 'rack', 'user', 'comment')
+        fields = ('hostname', 'elevation', 'model', 'rack', 'owner', 'comment')
```
juliasaveliff: Will be adding more serializers with future API PRs

rackcity/admin.py
clairecfu: woo goodBYE articles

rackcity/migrations/0002_itinstance_itmodel_rack_user.py
```diff
@@ -31,7 +33,8 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='Rack',
             fields=[
-                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('id', models.AutoField(auto_created=True,
```
clairecfu: is this the rid? / will it be sent with the other parameters?

juliasaveliff: @clairecfu `id` is an auto-incrementing field that Django adds to every row in a table. we are using that as the primary key, and it will be serialized in all GET requests as `rid` to be consistent with front end naming. didn't want to make a redundant `rid` field since it would serve exactly the same purpose

PR #29, opened 2020-01-26
added auth and validation to the models/add endpoint. Might be a good pattern for how to do validation

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -18,22 +18,31 @@ def model_list(request):
...
-            return HttpResponse(status=400)
+            return JsonResponse({"success": True})
+        except Exception as error:
+            failure_message = failure_message + str(error)
+
+    failure_message = "Request was invalid. " + failure_message
+    return JsonResponse({
+        "success": False,
+        "failure_message": failure_message
+    })
```
juliasaveliff: **_per our discussion_** I think it would be worthwhile to populate the `status` param on the `JsonResponse` with the corresponding HTTP status code, so we return something like: 

```
JsonResponse(
    {"error_description": "blah"},
    status=400,
)
``` 
on bad requests and 
```
 JsonResponse(
    {},
    status=200,
)
```
on good requests?

juliasaveliff: I think we should use these https://docs.python.org/3/library/http.html#http.HTTPStatus for better readability

benauriemma: Sick I'm down

rackcity/views/it_model_views.py
```diff
@@ -18,22 +18,33 @@ def model_list(request):
...
+    if failure_message == "":
+        try:
             serializer.save()
-            # should return something better maybe
-            return HttpResponse(status=201)
-        else:
-            print("invalid")
-            # should return something better definitely
-            return HttpResponse(status=400)
+            return JsonResponse({"success": True}, status=201)
```
benauriemma: I made this 201 "Created" but we can change it if it's difficult for front end

rackcity/views/it_model_views.py
```diff
@@ -18,22 +18,33 @@ def model_list(request):
...
     """
-    if request.method == 'POST':
-        data = JSONParser().parse(request)
-        serializer = ITModelSerializer(data=data)
-        if serializer.is_valid():  # TODO: figure out what .valid() does.  Different fields are required for add, update, etc
-            print("valid")
+    data = JSONParser().parse(request)
+    failure_message = ""
+    if 'id' in data:
+        failure_message = failure_message + "Don't include id when adding a model. "
```
juliasaveliff: do we want the request to fail if `id` is passed as a param? your current implementation allows this to continue on to the serialize attempt

rackcity/views/it_model_views.py
juliasaveliff: this may be too picky of me but I think we should use python's http library (`from http import HTTPStatus`) then replace `status=201` with `status=HTTPStatus.CREATED`

I think that makes the code more readable instead of having to remember the nuances between various 2xx, 4xx, 5xx errors

juliasaveliff: another thing - do we still need "success" in the response data or is that redundant? I think we could just send empty data dict for 200s

rackcity/views/it_model_views.py
benauriemma: So it would try to validate the object which I think is good because you'll get a more complete error message if there are other problems, but then it won't try to save the model because the error message is nonempty

rackcity/views/it_model_views.py
benauriemma: I changed this to be an HttpResponse since it doesn't have a body

juliasaveliff:
looks good!

PR #30, opened 2020-01-26
Implement simple APIs on instance

juliasaveliff:
- Implements `/instances/get-many`, sorting and filtering not functional yet
- Implements `/instances/add`


rackcity/api/serializers/it_instance_serializers.py
```diff
@@ -1,8 +1,35 @@
 from rest_framework import serializers
 from rackcity.models import ITInstance
+from .it_model_serializers import ITModelSerializer
+from .rack_serializers import RackSerializer
 
 
 class ITInstanceSerializer(serializers.ModelSerializer):
+    """
+    Serializes all fields on ITInstance model, where model and rack fields are
+    defined by their pk only.
```
juliasaveliff: want this for /add. don't want to have to pass in entirely defined model when creating a new instance, should only have to specify pk of model

rackcity/api/urls.py
```diff
@@ -8,7 +8,9 @@
     path('models/test-auth', views.model_auth),
     path('models/test-admin', views.model_admin),
     path('instances', views.instance_list),
```
juliasaveliff: to be deprecated ^^ keeping for now

rackcity/views/it_instance_views.py
```diff
@@ -1,19 +1,48 @@
...
+from rest_framework.decorators import permission_classes, api_view
+from rest_framework.permissions import IsAuthenticated, IsAdminUser
+from rest_framework.parsers import JSONParser
+from rest_framework.pagination import PageNumberPagination
+from http import HTTPStatus
 
 
+@api_view(['GET'])
+@permission_classes([IsAuthenticated])
 def instance_list(request):
```
benauriemma: This will be deprecated eventually right? Fine with it for now though

rackcity/views/it_instance_views.py
```diff
@@ -1,19 +1,48 @@
...
+@api_view(['POST'])
+@permission_classes([IsAuthenticated])
+def instance_page(request):
+    """
+    List a page of instances.
+    """
+    instances = ITInstance.objects.all()
+    paginator = PageNumberPagination()
+    paginator.page_size = 10
+    page_of_instances = paginator.paginate_queryset(instances, request)
```
benauriemma: I wish it was more clear that this gets the page number that was specified in the request but I'm not sure how to do that. Do you think it's worth an inline comment?

rackcity/views/it_instance_views.py
```diff
@@ -1,19 +1,48 @@
...
+    serializer = RecursiveITInstanceSerializer(page_of_instances, many=True)
+    return JsonResponse(
+        serializer.data,
+        safe=False,
```
benauriemma: What does safe=False mean?

rackcity/views/it_instance_views.py
```diff
@@ -24,5 +53,34 @@ def instance_detail(request, pk):
...
 
     if request.method == 'GET':
-        serializer = ITInstanceSerializer(instance)
+        serializer = RecursiveITInstanceSerializer(instance)
         return JsonResponse(serializer.data)
+
+
+@api_view(['POST'])
+@permission_classes([IsAdminUser])
+def instance_add(request):
```
benauriemma: Lovely

rackcity/views/it_instance_views.py
```diff
@@ -24,5 +53,34 @@ def instance_detail(request, pk):
         return HttpResponse(status=404)
```
benauriemma: Is this endpoint considered done or WIP? If it's done I have a couple comments

rackcity/views/it_instance_views.py
juliasaveliff: Indeed. Keeping it so we don't break existing endpoints..........yet

rackcity/views/it_instance_views.py
juliasaveliff: WIP

rackcity/views/it_instance_views.py
juliasaveliff: I agree. Will add comment for now

PR #33, opened 2020-01-26
models/id endpoint

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -47,6 +47,27 @@ def model_add(request):
...
+@permission_classes([IsAuthenticated])
+def model_detail(request, id):
+    """
+    Retrieve a single model.
+    """
+    try:
+        model = ITModel.objects.get(id=id)
+        model_serializer = ITModelSerializer(model)
+        instances = ITInstance.objects.filter(model=id)
+        instances_serializer = ITInstanceSerializer(instances, many=True)
```
benauriemma: I didn't use the Recursive one here because we don't need the model info repeated, right?

juliasaveliff: agreed

rackcity/views/it_model_views.py
```diff
@@ -47,6 +47,27 @@ def model_add(request):
...
+        model_detail = {
+            "model": model_serializer.data,
+            "instances": instances_serializer.data
+        }
+        return JsonResponse(model_detail, status=HTTPStatus.OK)
+    except ITModel.DoesNotExist:
```
juliasaveliff: I think it's best practice to keep minimal lines of code inside the try block (mostly for readability so it's clearer where the exception comes from), if you're handling your except block such that the rest of the code path would never be hit

imma approve this and leave that up to you though, not a huge deal since this exception is pretty obvious about what went wrong

benauriemma: Cool will keep in mind in the future

PR #39, opened 2020-01-27
implemented models/modify. a couple parts of this are a little bit funky but it do be working

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -48,6 +49,46 @@ def model_add(request):
...
+            ITModel.objects.get(id=id)
+        except ObjectDoesNotExist:
+            failure_message += "No existing model with id="+str(id)+". "
+        else:
+            serializer = ITModelSerializer(data=data)
+            if not serializer.is_valid(raise_exception=False):
+                failure_message += str(serializer.errors)
+            if failure_message == "":
+                existing_model = ITModel.objects.get(id=id)
+                print(serializer.data)
```
juliasaveliff: should remove this print before merging

rackcity/views/it_model_views.py
```diff
@@ -48,6 +49,46 @@ def model_add(request):
...
+                for field in serializer.data.keys():
+                    if serializer.data[field] is not None:
+                        setattr(existing_model, field,
+                                serializer.data[field])
+                try:
+                    existing_model.save()
+                    return HttpResponse(status=HTTPStatus.CREATED)
```
juliasaveliff: technically nothing was CREATED? maybe use OK instead?

rackcity/views/it_model_views.py
```diff
@@ -48,6 +49,46 @@ def model_add(request):
...
+        failure_message += "Must include id when modifying a model. "
+    else:
+        id = data['id']
+        try:
+            ITModel.objects.get(id=id)
+        except ObjectDoesNotExist:
+            failure_message += "No existing model with id="+str(id)+". "
+        else:
+            serializer = ITModelSerializer(data=data)
+            if not serializer.is_valid(raise_exception=False):
```
juliasaveliff: This is going to fail if the required model objects are not included (vendor, model number, height). We could either make a new Serializer for which all fields are not required (personally I think this is the better option), or we could implement deserialization ourselves just checking each `serializer.data[field] is not None` and then validating the value of each...!

PR #40, opened 2020-01-27
implemented models/delete endpoint

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -48,6 +49,38 @@ def model_add(request):
...
+        id = data['id']
+        try:
+            ITModel.objects.get(id=id)
+        except ObjectDoesNotExist:
+            failure_message += "No existing model with id="+str(id)+". "
+        else:
+            if failure_message == "":
+                existing_model = ITModel.objects.get(id=id)
+                try:
+                    existing_model.delete()
```
juliasaveliff: Not sure how `.delete()` is implemented. We don't want to delete the model if any instances of this model type exist. Will this fail if instances with the model exist because of the `ForeignKey` relationship?

juliasaveliff: Also might need to discuss this further in person but the requirements state the user has to confirm on a delete. Maybe front end would do a GET on model, see if any instances exist, if so don't even call the delete endpoint and tell the user they can't delete then?

benauriemma: Wow I fully had it in my head that I would make this check for instances before deleting and then just... didn't

benauriemma: So you need confirmation from the user no matter what, but then only are allowed to do the delete if no instances exist. I was picturing the front end would have a "dumb" popup that says are you sure?? before sending the request at all, and then the request assumes confirmation was given but fails if there are instances that exist

juliasaveliff:
```
"The administrator seeking to remove a model shall only be able to do so if no
instances of the model exist. Even in this case, a confirmation dialog shall be
shown."
```

yeah I guess you're right. but there should be some popup if the delete fails explaining that instances of it exist? or will we just say "removal failed"?

benauriemma: Yep! Now that it checks for that the error message will include `"Cannot delete this model because instances of it exist. "` if it's an issue

PR #41, opened 2020-01-27
Implements /racks/get endpoint

juliasaveliff:
TO the best of my knowledge, this is the best way I could figure out how to take advantage of Serializer class for deserializing non-restful POST body, while avoiding blindly checking for string fields in a data dictionary in the function view

Open to commentary but I realized I said best twice in this description so I am either (1) certainly the best or (2) very tired


rackcity/api/objects/rack_range.py
```diff
@@ -0,0 +1,44 @@
+from rest_framework import serializers
+
+
+class RackRangeSerializer(serializers.Serializer):
+    letter_start = serializers.CharField(required=True, max_length=1)
+    letter_end = serializers.CharField(required=False, max_length=1)
+    num_start = serializers.IntegerField(required=True)
+    num_end = serializers.IntegerField(required=False)
+
+    def validate(self, data):
```
juliasaveliff: The serializer `.is_valid()` method checks only that the JSON body has fields of the correct type and arguments (max_length, required). Added validation with this method to enforce that END fields are greater than START fields.

rackcity/api/objects/rack_range.py
```diff
@@ -0,0 +1,44 @@
...
+        if ('letter_end' in data) and (
+            data['letter_end'] < data['letter_start']
+        ):
+            message = "invalid rack row range from " + \
+                data['letter_start'] + " to " + data['letter_end']
+            raise serializers.ValidationError(message)
+
+        return data
+
+    def get_row_range(self, data):
```
juliasaveliff: Django will not allow these methods to be called BEFORE an `.is_valid()` check is called (how Serializer class is implemented), so guaranteed that at least the start fields will exist on data

rackcity/views/rack_views.py
```diff
@@ -26,3 +30,66 @@ def rack_detail(request, pk):
...
+@api_view(['POST'])
+@permission_classes([IsAuthenticated])
+def rack_get(request):
+    """
+    List all racks within specified range.
+    """
+    range_serializer = RackRangeSerializer(data=request.data)
+    if range_serializer.is_valid():
+        try:
+            range_serializer.validate(request.data)
```
juliasaveliff: Not the most beautiful order of of calls. But Django requires check on `.is_valid()` before any other methods can be called on the serializer object

rackcity/views/rack_views.py
```diff
@@ -26,3 +30,66 @@ def rack_detail(request, pk):
...
+            )
+    else:
+        return JsonResponse(
+            {"failure_message": "invalid keys or values"},
+            status=HTTPStatus.BAD_REQUEST,
+        )
+
+    racks = Rack.objects.filter(
+        rack_num__range=range_serializer.get_number_range(request.data),
+        row_letter__range=range_serializer.get_row_range(request.data),
```
juliasaveliff: The `__range` filter is inclusive on both max and min

rackcity/api/objects/rack_range.py
benauriemma: I also thought about calling is_valid() within validate(), but I'm not sure that would be allowed since is_valid() is supposed to be the first thing called?  If that's allowed I think it would be cleaner

rackcity/views/rack_views.py
benauriemma: Worth including as an inline comment?

rackcity/views/rack_views.py
```diff
@@ -26,3 +30,66 @@ def rack_detail(request, pk):
...
+
+    return JsonResponse(
+        {"racks": racks_with_instances},
+        status=HTTPStatus.OK
+    )
+
+
+@api_view(['POST'])
+@permission_classes([IsAdminUser])
+def rack_create(request):
```
benauriemma: Sorry to be picky but can we make these return `JsonResponse({}, status=5xx)` until they're implemented in case Brendan or Claire aren't tracking

rackcity/api/objects/rack_range.py
juliasaveliff: Yeah I tried that too and it is not allowed (mentioned that in the other comment below)

`validate()` is a method on the `Serializer` class that I'm overriding. I just did some more investigating and this `validate()` method is the last thing that's called from `is_valid()`, so we don't need to do both here. The only bummer about that is only one validation exception is ever thrown from `is_valid()` so we couldn't specify which part of the request went wrong in the response. But I don't think that's a huge problem for now

Will update this PR with my changeszsszsszs

rackcity/views/rack_views.py
juliasaveliff: done

rackcity/api/objects/rack_range.py
juliasaveliff: Yeah I tried that too and it is not allowed (mentioned that in the other comment below)

`validate()` is a method on the `Serializer` class that I'm overriding. I just did some more investigating and this `validate()` method is the last thing that's called from `is_valid()`, so we don't need to do both here. The only bummer about that is only one validation exception is ever thrown from `is_valid()` so we couldn't specify which part of the request went wrong in the response. But I don't think that's a huge problem for now

Will update this PR with my changeszsszsszs

rackcity/views/rack_views.py
```diff
@@ -26,3 +30,66 @@ def rack_detail(request, pk):
...
+    List all racks within specified range.
+    """
+    range_serializer = RackRangeSerializer(data=request.data)
+    if range_serializer.is_valid():
+        try:
+            range_serializer.validate(request.data)
+        except Exception as error:
+            return JsonResponse(
+                {"failure_message": str(error)},
+                status=HTTPStatus.BAD_REQUEST
```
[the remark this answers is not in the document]
juliasaveliff: Oh I forgot about that. Although NOT_ACCEPTABLE has a nice ring to it, I don't think that's correct because that indicates the request did NOT necessarily fail whereas BAD_REQUEST status indicates bad syntax & that the "client SHOULD NOT repeat the request without modifications"

rackcity/views/rack_views.py
```diff
@@ -26,3 +30,59 @@ def rack_detail(request, pk):
...
+
+@api_view(['POST'])
+@permission_classes([IsAuthenticated])
+def rack_get(request):
+    """
+    List all racks within specified range.
+    """
+    range_serializer = RackRangeSerializer(data=request.data)
+
+    if not range_serializer.is_valid():
```
juliasaveliff: `is_valid()` calls `validate()` method implemented on RackRangeSerializer class

rackcity/api/objects/rack_range.py
juliasaveliff: Accidentally resolved these comments....I have pushed related updates!

PR #44, opened 2020-01-28
Create Model Form + BIG table refactor 

clairecfu:
Refactored Table Component Classes, as well as defined interfaces for ModelObject,InstanceObject, and RackObjects. These and other useful functions live in the `utils.tsx` folder.

The primary goal of this PR is to standardize types across the project, and also create a 1:1 mapping of the Model/Rack/Instance Objects returned in the API endpoint to Typescript Interfaces


src/components/elementView/elementView.tsx
```diff
@@ -1,143 +1,98 @@
...
-  data: any;
-}
-interface ElementTableProps {
-  element: string;
-  history: any;
-}
+function getElementData(
+  path: string,
+  token: string
+): Promise<Array<ElementObjectType>> {
```
clairecfu: Creating object types allows me to directly map Axios response objects to interfaces that we defined.

src/components/utils.tsx
```diff
@@ -0,0 +1,45 @@
...
+export type ElementObjectType = ModelObject | RackObject | InstanceObject;
+export function isModelObject(obj: any): obj is ModelObject {
+  return obj && obj.model_number;
+}
+export function isRackObject(obj: any): obj is RackObject {
+  return obj && obj.rack_num;
+}
+export function isInstanceObject(obj: any): obj is InstanceObject {
+  return obj && obj.hostname;
+}
```
clairecfu: This is the most important part of the PR to look at !! Typescript does not have the traditional objects found in java, but it DOES support objects that extend certain interfaces. This is the most useful for mapping axios response objects into objects that extend these interfaces. Also added three functions that can be helpful to check the type of object that you are working with.

src/App.tsx
```diff
@@ -55,7 +56,7 @@ class App extends React.Component<AppProps> {
   }
 }
 
-const mapStateToProps = (state: any) => {
+const mapStateToProps = (state: any, ownProps: any = {}) => {
```
brendancheng20: Is `ownProps: any = {}` something that needs to be included in components that use mapStateToProps?

clairecfu:
oops no this was from when i was testing stuff out, will remove

src/components/elementView/elementTabView.tsx
```diff
@@ -0,0 +1,53 @@
...
+        /> */}
+        <Tabs.Expander />
+      </Tabs>
+    );
+  }
+}
+
+const mapStateToProps = (state: any) => {
+  return {
+    isAdmin: state.admin
```
brendancheng20: Similar question to before, but why does this mapStatetoProps not have the usual `isAuthenticated: state.token !== null` ?

clairecfu: you can choose to map any set of redux states to any arbitrarily named properties, elementTabView doesn't need to know if the user is Authenticated because its parent container (elementTab) already handles the logic

src/components/elementView/elementView.tsx
brendancheng20: To be clear, does this just specify the return type of the function so that you can "access" data from it before the Promise is resolved?

src/components/utils.tsx
brendancheng20: Can you explain how the functions work? I get how the type predicating works with `obj is ________` but what's going on with the return values?

src/App.tsx
clairecfu: i tried to respond to this via email but it just left a comment on the PR so incase you missed it "oops no this was from when i was testing stuff out, will remove", i removed this in a new commit

src/components/utils.tsx
clairecfu: The return properties are kind of arbitrary, but the idea is that im making the assumptions model_number is a unique property ModelObject, so that any object with that property is a ModelObject, so if the property exists, the function will return true. This could be modified with more properties if necessary though

src/components/elementView/elementView.tsx
clairecfu: the return type of the function is still in a Promise, mainly because I couldn't figure out how to truly get it out of the Promise function to pass into a diff function. BUT, i basically just take this promise object and in my ElementTable I call ```    this.props.getData(this.props.type, this.props.token).then(res => {
      this.setState({
        items: res
      });
    });``` to truly get the value out of the promise.

clairecfu: Also specifying that the elements inside the Promise are an array of objects of type`ElementObjectType`, which the Table needs

PR #46, opened 2020-01-28
Implemented racks/create, racks/delete

juliasaveliff:
Following implementation decisions were made:
- On create:
   - If some/all racks within specified range already exist, request fails and no changes are made (does NOT specify which racks exist)
- On delete:
   - If some/all racks within specified range do not exist, request fails and no changes are made (does NOT specify which racks do not exist) 
   - If some/all racks within specified range have instances in them, request fails and no changes are made (specifies which racks have instances in them) 

Subject to change according to response at: https://piazza.com/class/k4u27qnccr45oo?cid=50

### DIFF
--- a/rackcity/views/rack_views.py
+++ b/rackcity/views/rack_views.py
@@ -42,7 +42,7 @@ def rack_get(request):
 
     if not range_serializer.is_valid():
         return JsonResponse(
-            {"failure_message": "Invalid keys or values in request."},
+            {"failure_message": str(range_serializer.errors)},
             status=HTTPStatus.BAD_REQUEST,
         )
 
@@ -79,7 +79,7 @@ def rack_create(request):
     range_serializer = RackRangeSerializer(data=request.data)
     if not range_serializer.is_valid():
         return JsonResponse(
-            {"failure_message": "Invalid keys or values in request."},
+            {"failure_message": str(range_serializer.errors)},
             status=HTTPStatus.BAD_REQUEST,
         )
     racks = Rack.objects.filter(
@@ -112,7 +112,7 @@ def rack_delete(request):
     range_serializer = RackRangeSerializer(data=request.data)
     if not range_serializer.is_valid():
         return JsonResponse(
-            {"failure_message": "Invalid keys or values in request."},
+            {"failure_message": str(range_serializer.errors)},
             status=HTTPStatus.BAD_REQUEST,
         )
     racks = Rack.objects.filter(
--- a/rackcity/views/rack_views.py
+++ b/rackcity/views/rack_views.py
@@ -76,7 +76,31 @@ def rack_create(request):
     """
     Create racks within specified range.
     """
-    return JsonResponse({}, status=HTTPStatus.INTERNAL_SERVER_ERROR)
+    range_serializer = RackRangeSerializer(data=request.data)
+    if not range_serializer.is_valid():
+        return JsonResponse(
+            {"failure_message": "Invalid keys or values in request."},
+            status=HTTPStatus.BAD_REQUEST,
+        )
+    racks = Rack.objects.filter(
+        rack_num__range=range_serializer.get_number_range(),  # inclusive range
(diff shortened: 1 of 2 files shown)


benauriemma:
You said "If some/all racks within specified range do not exist, request fails and no changes are made (does NOT specify which racks do not exist)" but I thought we were returning a string with info about that?

benauriemma:
Did/can you document those decisions somewhere for posterity?

rackcity/views/rack_views.py
```diff
@@ -76,7 +76,31 @@ def rack_create(request):
     """
     Create racks within specified range.
     """
-    return JsonResponse({}, status=HTTPStatus.INTERNAL_SERVER_ERROR)
+    range_serializer = RackRangeSerializer(data=request.data)
+    if not range_serializer.is_valid():
+        return JsonResponse(
+            {"failure_message": "Invalid keys or values in request."},
```
benauriemma: I think you can do `{"failure_message": str(range_serializer.errors)}` here can't you?

rackcity/views/rack_views.py
```diff
@@ -85,4 +109,48 @@ def rack_delete(request):
     """
     Delete racks within specified range.
     """
-    return JsonResponse({}, status=HTTPStatus.INTERNAL_SERVER_ERROR)
+    range_serializer = RackRangeSerializer(data=request.data)
+    if not range_serializer.is_valid():
+        return JsonResponse(
+            {"failure_message": "Invalid keys or values in request."},
```
benauriemma: Same as above

rackcity/views/rack_views.py
```diff
@@ -85,4 +109,48 @@ def rack_delete(request):
...
+    if len(unempty_racks) > 0:
+        failure_message = "No racks were deleted because the following " + \
+            "racks contain instances: " + ", ".join(unempty_racks)
+        return JsonResponse(
+            {"failure_message": failure_message},
+            status=HTTPStatus.BAD_REQUEST,
+        )
+    else:
+        try:
+            racks.delete()
```
benauriemma: You can call this on a list of models? Wild

rackcity/views/rack_views.py
```diff
@@ -76,7 +76,31 @@ def rack_create(request):
...
+            status=HTTPStatus.BAD_REQUEST,
+        )
+    racks = Rack.objects.filter(
+        rack_num__range=range_serializer.get_number_range(),  # inclusive range
+        row_letter__range=range_serializer.get_row_range(),
+    )
+    if racks.count() > 0:
+        failure_message = "Racks cannot be created because racks within " + \
+            "the range already exist."
+        return JsonResponse(
```
benauriemma: Seems like you're following the fail early pattern here which we don't really do elsewhere. I don't dislike it but worth tracking and considering standardizing one way or the other at some point down the line

benauriemma:
How crazy is it that we did vooga with like 4 times as many people and were definitely was worse engineers at the time but simply had no code review protocol? Like you could really just merge whatever you wanted straight to master

rackcity/views/rack_views.py
juliasaveliff: so true. fixing this

rackcity/views/rack_views.py
juliasaveliff: p what? p wild

rackcity/views/rack_views.py
juliasaveliff: tracking engaged

juliasaveliff:
@benauriemma aDdEd cHanGeS cAn YOu pLeASe rEviEW aNd apProvE

juliasaveliff:
Milestone: I just accidentally closed a PR

rackcity/views/rack_views.py
benauriemma: Bippity bump

benauriemma: Wait I take it back I think you fixed this

PR #48, opened 2020-01-28
Fix instances add (checks rack location) 

juliasaveliff:
- TODO (on diff branch) implement instances modify with this check too

### DIFF
--- a/rackcity/views/it_instance_views.py
+++ b/rackcity/views/it_instance_views.py
@@ -1,6 +1,7 @@
 # from rest_framework.parsers import JSONParser
 from django.http import HttpResponse, JsonResponse
 from rackcity.models import ITInstance, ITModel, Rack
+from django.core.exceptions import ObjectDoesNotExist
 from rackcity.api.serializers import (
     ITInstanceSerializer,
     RecursiveITInstanceSerializer,
@@ -13,7 +14,7 @@ from http import HTTPStatus
 import math
 
 
-@api_view(['GET'])
+@api_view(['GET'])  # DEPRECATED !
 @permission_classes([IsAuthenticated])
 def instance_list(request):
     """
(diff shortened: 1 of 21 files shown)


rackcity/views/it_instance_views.py
```diff
@@ -167,3 +174,15 @@ def instance_page_count(request):
     instance_count = ITInstance.objects.all().count()
     page_count = math.ceil(instance_count / page_size)
     return JsonResponse({"page_count": page_count})
+
+
+def is_location_full(rack_id, elevation, height):
```
benauriemma: I think it might be helpful to name this `instance_height` rather than `height`, I originally wasn't sure if it was instance height or rack height

rackcity/views/it_instance_views.py
```diff
@@ -167,3 +174,15 @@ def instance_page_count(request):
...
+    rack_u = [elevation + i for i in range(height)]
```
benauriemma: I think this could be named better too - maybe `new_instance_location_range` or something like that?

benauriemma: That's kind of clunky I admit but I think rack_u is unclear

rackcity/views/it_instance_views.py
```diff
@@ -167,3 +174,15 @@ def instance_page_count(request):
...
+    instances_in_rack = ITInstance.objects.filter(rack=rack_id)
+    for instance in instances_in_rack:
+        for u in [
```
benauriemma: Same comment for `u`. Maybe `occupied_range`?

benauriemma: Actually I guess `occupied_slot` or `occupied_location` is more accurate

rackcity/views/it_instance_views.py
juliasaveliff: on it

rackcity/views/it_instance_views.py
juliasaveliff: I like

rackcity/views/it_instance_views.py
juliasaveliff: excellent

juliasaveliff:
@benauriemma updated variable names 👀

PR #51, opened 2020-01-28
Bauriemma models bulk upload endpoint

benauriemma:
Tested by hand and it looks good but definitely need to do more testing with files, etc.

Also need to implement the approval endpoint.

If it looks good, though, thoughts on merging so we can start using it to tease out bugs and push front end along?

### DIFF
--- a/rackcity/views/it_model_views.py
+++ b/rackcity/views/it_model_views.py
@@ -7,6 +7,7 @@ from rest_framework.decorators import permission_classes, api_view
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.pagination import PageNumberPagination
 from http import HTTPStatus
+from copy import copy
 import math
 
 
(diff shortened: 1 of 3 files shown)


rackcity/views/it_model_views.py
```diff
@@ -194,6 +195,82 @@ def model_vendors(request):
...
+        },
+        status=HTTPStatus.OK
+    )
+
+
+def records_are_identical(existing_data, new_data):
+    existing_keys = existing_data.keys()
+    new_keys = new_data.keys()
+    for key in existing_keys:
+        if key not in new_keys and existing_data[key] is not None and key != 'id':
```
juliasaveliff: If key is an optional field (such as "cpu") and the "cpu" field on the existing model is NULL, did you double check that `data["cpu"] = None` in this case? that's exactly what I would guess happens, but want to make sure we checked this case in case Django is doing something unexpected with NULL fields

benauriemma: Yep I tested and if existing model has:
`"storage": null,`
and the new model has either 
`"storage": null,`
or does not have a key `"storage"`, they're considered equivalent

PR #54, opened 2020-01-29
models bulk approve endpoint

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -271,6 +273,35 @@ def records_are_identical(existing_data, new_data):
...
+            {"failure_message": "Bulk approve request should have a parameter 'approved_modifications'"},
+            status=HTTPStatus.BAD_REQUEST
+        )
+    model_datas = data['approved_modifications']
+    for model_data in model_datas:
+        model_serializer = ITModelSerializer(data=model_data)
+        if not model_serializer.is_valid():
+            failure_message = "At least one modification was not valid. " + \
+                str(model_serializer.errors)
+            return JsonResponse(
```
juliasaveliff: Let's say the user sends 3 approved modifications. The 1st and 3rd are valid and the 2nd is not. In this case, the 1st modification will be saved on the existing model. When we detect the 2nd one is invalid, the request fails. That means the 3rd valid modification never gets updated. 

Are we just assuming something like this will never happen, since if our API is followed correctly, all of these modifications _should_ have already been validated on bulk-upload? Or should we be iterating over all `model_data` to validate first, and then do a second iteration to save all models?

benauriemma: My thought was that this should never happen because validity was checked before, and these models should be given back to us exactly how they were before but I'm down to do all the checks first then all the saves just to be safe. I'll rip the change

PR #56, opened 2020-01-29
instances returned from racks/get now have the full model included, theyre also sorted by elevation

benauriemma:
rack is also included with instance which isnt ideal but whatev

### DIFF
--- a/rackcity/views/rack_views.py
+++ b/rackcity/views/rack_views.py
@@ -1,7 +1,7 @@
 # from rest_framework.parsers import JSONParser
 from django.http import HttpResponse, JsonResponse
 from rackcity.models import Rack, ITInstance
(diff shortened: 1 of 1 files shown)


rackcity/views/rack_views.py
```diff
@@ -29,8 +29,10 @@ def rack_get(request):
     racks_with_instances = []
     for rack in racks:
         rack_serializer = RackSerializer(rack)
-        instances = ITInstance.objects.filter(rack=rack.id)
-        instances_serializer = ITInstanceSerializer(
+        instances = ITInstance.objects \
+            .filter(rack=rack.id) \
+            .order_by("elevation")
```
clairecfu: what a handy method!

juliasaveliff:
I've laid my eyes on this. Delete away!

PR #57, opened 2020-01-29
bulk upload considers str(8) the same as 8

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -267,7 +267,8 @@ def records_are_identical(existing_data, new_data):
         if key not in new_keys and existing_data[key] is not None and key != 'id':
```
juliasaveliff: I'm worried about these conditionals getting hard to track. Could you possibly add some parentheses between the `and`s so it's a little more readable?

rackcity/views/it_model_views.py
```diff
@@ -267,7 +267,8 @@ def records_are_identical(existing_data, new_data):
...
             return False
         if key in new_keys and new_data[key] != existing_data[key]:
-            return False
+            if not (isinstance(existing_data[key], int) and int(new_data[key]) == existing_data[key]):
```
juliasaveliff: should this be `if not (isinstance(new_data[key], int)` ? because we're worried about data from the front end being sent as strings? 

So if `new_data[key] = "8"` and `existing_data[key] = 8`, that's a match
But if `new_data[key] = "8" and `existing_data[key] = 9, that's not a match

juliasaveliff: scratch that never mind, I was reading the `(` wrong 🥴
thoughts on splitting this into multiple lines so it's more readable?

benauriemma: Yeppo will try to make more readable

rackcity/views/it_model_views.py
benauriemma: Ditto

PR #58, opened 2020-01-30
Adding filter to get-many on instances

juliasaveliff:
For instances: 
- Separate sorting logic 
- Implement filtering (except for on rack range)


rackcity/views/it_instance_views.py
```diff
@@ -205,3 +208,72 @@ def is_location_full(rack_id, instance_elevation, instance_height):
...
+                elif filter_dict['match_type'] == 'contains':
+                    filter_args['{0}__icontains'.format(filter_field)] = \
+                        filter_dict['value']
+
+            elif filter_type == 'numeric':
+                range_value = (
+                    int(filter_dict['min']),
+                    int(filter_dict['max'])
+                )
+                filter_args['{0}__range'.format(filter_field)] = range_value
```
benauriemma: Can you leave a comment about whether this is inclusive on min/max?

rackcity/views/it_instance_views.py
```diff
@@ -205,3 +208,72 @@ def is_location_full(rack_id, instance_elevation, instance_height):
...
+
+            elif filter_type == 'rack_range':
```
benauriemma: Do we have a ticket to add this in later?

juliasaveliff: just made one!

rackcity/views/it_instance_views.py
juliasaveliff: adding comment

PR #62, opened 2020-01-30
Implement /instances/modify with location validation

juliasaveliff:
Tested locally with following cases
- Move instance to conflicting spot in same rack
- Move instance to non-conflicting spot in same rack, but also change model (diff. height), so spot becomes conflicting
- Move instance to conflicting spot in same rack, but also change model (diff. height), so spot is no longer conflicting
- Move instance to different spot in different rack, which is conflicting

### DIFF
--- a/rackcity/views/it_instance_views.py
+++ b/rackcity/views/it_instance_views.py
@@ -259,10 +259,7 @@ def is_location_full(
     instances_in_rack = ITInstance.objects.filter(rack=rack_id)
     for instance_in_rack in instances_in_rack:
         # Ignore if instance being modified conflicts with its old location
-        if (
-            (instance_id is None)
-            or (instance_id is not None and instance_in_rack.id != instance_id)
-        ):
+        if (instance_id is None or instance_in_rack.id != instance_id):
             for occupied_location in [
                 instance_in_rack.elevation + i for i
                     in range(instance_in_rack.model.height)
--- a/rackcity/views/it_instance_views.py
+++ b/rackcity/views/it_instance_views.py
@@ -123,12 +123,11 @@ def instance_add(request):
     if not serializer.is_valid(raise_exception=False):
         failure_message += str(serializer.errors)
 
-    instance_id = data['id']
     rack_id = data['rack']
     elevation = data['elevation']
     height = ITModel.objects.get(id=data['model']).height
 
-    if is_location_full(rack_id, instance_id, elevation, height):
+    if is_location_full(rack_id, elevation, height):
         failure_message += "Instance does not fit in this location. "
 
     if failure_message == "":
@@ -250,9 +249,9 @@ def instance_page_count(request):
 
 def is_location_full(
     rack_id,
-    instance_id,
     instance_elevation,
     instance_height,
+    instance_id=None,
 ):
     new_instance_location_range = [
         instance_elevation + i for i in range(instance_height)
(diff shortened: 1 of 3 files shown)


rackcity/views/it_instance_views.py
```diff
@@ -141,6 +141,58 @@ def instance_add(request):
...
+        )
+
+    for field in data.keys():
+        if field == 'model':
+            value = ITModel.objects.get(id=data[field])
+        elif field == 'rack':
+            value = Rack.objects.get(id=data[field])
+        else:
+            value = data[field]
+        setattr(existing_instance, field, value)
```
juliasaveliff: Using `setattr` requires entire model and rack objects, can't just set by id

rackcity/views/it_instance_views.py
```diff
@@ -192,16 +244,53 @@ def instance_page_count(request):
...
+            for occupied_location in [
+                instance_in_rack.elevation + i for i
+                    in range(instance_in_rack.model.height)
+            ]:
+                if occupied_location in new_instance_location_range:
+                    return True
     return False
+
+
+def validate_location_modification(data, existing_instance):
```
benauriemma: What would you think about making this function agnostic to whether it's a modification or an addition of an instance? For example by getting rid of the `existing_instance` parameter and then within the function checking if `'id' in data` to determine if there's an existing instance.

I think it would be generally nice to have the validation be the same for both cases, and it would be especially helpful for bulk import because additions and modifications happen all together and otherwise undergo the same validation logic.

I'm not really sure about this so feel free to dissent

juliasaveliff: while I'm all about reducing duplicated code where we can, the difference between the two cases right now is that we use the serializer to validate incoming JSON data for `/instances/add` endpoint, and we can't use it for modifications because of the required field issue. I don't think this is worth getting rid of the serializer on adds because it's a more thorough validation. I think in the future we'll definitely want to move towards a more universal type of validation on each model, and I'm not sure what that'll look like yet, but I'm hesitant about throwing it all into this `validate_location_modification` method

thoughts???

benauriemma: I meant making this method which is just for validating location agnostic to add vs modification, not necessarily the rest of the validation logic like for fields and uniqueness.  But if you disagree then I don't mind approving it

rackcity/views/it_instance_views.py
```diff
@@ -195,21 +247,66 @@ def instance_page_count(request):
...
-        for occupied_location in [
-            instance_in_rack.elevation + i for i
-                in range(instance_in_rack.model.height)
-        ]:
-            if occupied_location in new_instance_location_range:
-                return True
+        # Ignore if instance being modified conflicts with its old location
+        if (
+            (instance_id is None)
+            or (instance_id is not None and instance_in_rack.id != instance_id)
```
juliasaveliff: @benauriemma  fixed bug from before

rackcity/views/it_instance_views.py
```diff
@@ -195,21 +247,63 @@ def instance_page_count(request):
...
     instances_in_rack = ITInstance.objects.filter(rack=rack_id)
     for instance_in_rack in instances_in_rack:
-        for occupied_location in [
-            instance_in_rack.elevation + i for i
-                in range(instance_in_rack.model.height)
-        ]:
-            if occupied_location in new_instance_location_range:
-                return True
+        # Ignore if instance being modified conflicts with its old location
+        if (instance_id is None or instance_in_rack.id != instance_id):
```
juliasaveliff: @benauriemma fixed bug from before

PR #66, opened 2020-01-30
Brendan bulk import

brendancheng20:
Model bulk import with modifications fully functional


package.json
```diff
@@ -3,33 +3,34 @@
...
     "@blueprintjs/icons": "^3.13.0",
     "@blueprintjs/select": "^3.11.2",
-    "@blueprintjs/table": "^3.8.2",
+    "@blueprintjs/table": "^3.8.3",
     "@blueprintjs/timezone": "^3.6.1",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.4.0",
     "@testing-library/user-event": "^7.2.1",
-    "@types/jest": "^24.9.0",
+    "@types/jest": "^24.9.1",
```
clairecfu: did these versions auto update? ideally we want to avoid doing this automatically because upgrading packages could break things. but if it didn't break anything then i guess were good for now lol

brendancheng20: Yeah they auto updated, didn't break anything when running locally tho. How can I avoid auto update in future?

PR #74, opened 2020-01-31
instances bulk location validation

benauriemma:
Just the basics

### DIFF
--- a/rackcity/views/it_instance_views.py
+++ b/rackcity/views/it_instance_views.py
@@ -314,14 +314,41 @@ def instance_bulk_upload(request):
             existing_instance = ITInstance.objects.get(
                 hostname=instance_data['hostname'])
         except ObjectDoesNotExist:
-            instances_to_add.append(instance_serializer)
+            model = ITModel.objects.get(id=instance_data['model'])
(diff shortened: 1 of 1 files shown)


rackcity/views/it_instance_views.py
```diff
@@ -310,14 +314,41 @@ def instance_bulk_upload(request):
             existing_instance = ITInstance.objects.get(
                 hostname=instance_data['hostname'])
         except ObjectDoesNotExist:
-            instances_to_add.append(instance_serializer)
+            model = ITModel.objects.get(id=instance_data['model'])
+            if is_location_full(
```
benauriemma: I still think it'd be nice to be able to use validate_location_modification (maybe by a different name) here. If you agree I'm down to change it

benauriemma:
Okay so ACTUALLY if an instance in the import file conflicts with an instance in the database that's not necessarily a dealbreaker... what if the one in the database gets moved by the import??

rackcity/views/it_instance_views.py
juliasaveliff: I agree

PR #75, opened 2020-01-31
models bulk location validation

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -241,6 +269,15 @@ def model_bulk_upload(request):
         except ObjectDoesNotExist:
             models_to_add.append(model_serializer)
         else:
+            if not model_height_change_valid(model_data, existing_model):
+                failure_message = \
+                    "Height change of this model causes conflicts: " + \
+                    "vendor="+model_data['vendor'] + \
+                    ", model_number="+model_data['model_number']
```
juliasaveliff: nice failure message

rackcity/views/it_model_views.py
```diff
@@ -83,6 +93,24 @@ def model_modify(request):
...
+def model_height_change_valid(new_model_data, existing_model):
+    if 'height' not in new_model_data:
+        return True
+    if new_model_data['height'] <= existing_model.height:
+        return True
+    else:
+        instances = ITInstance.objects.filter(model=existing_model.id)
+        for instance in instances:
+            if is_location_full(
+                instance.rack,
```
juliasaveliff: addressed this in the original PR, but make sure this is `instance.rack.id` when you merge

rackcity/views/it_model_views.py
benauriemma: ty

rackcity/views/it_model_views.py
benauriemma: Weird it seems to work either way but you are correct

PR #77, opened 2020-01-31
filter and sort on models/get-page

juliasaveliff:


rackcity/views/it_model_views.py
```diff
@@ -7,11 +7,12 @@
 from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from rest_framework.pagination import PageNumberPagination
 from http import HTTPStatus
-from copy import copy
```
benauriemma: Thanks I forgot about this

rackcity/views/it_model_views.py
```diff
@@ -169,22 +170,25 @@ def model_page(request):
...
+    except Exception as error:
+        return JsonResponse(
+            {"failure_message": "Filter error: " + str(error)},
+            status=HTTPStatus.BAD_REQUEST
+        )
+    models_query = models_query.filter(**filter_args)
+
+    try:
+        sort_args = get_sort_arguments(request.data)
+    except Exception as error:
```
benauriemma: I ran into a problem doing something like this where I got an unexpected exception (not the one I was intentionally raising in the method) because something went wrong in the method, but caught it thinking it was mine. Then I ended up displaying to the user something like "Sort error: couldn't find key 'id' in model_data"

I think better practice here is to make our own exception types and then say `except SortArgsException as error:`. Check out #81 for an example

Approving anyway bc that doesn't need to happen now

rackcity/views/it_model_views.py
juliasaveliff: hey no worries. pylint yelled at me.

rackcity/views/it_model_views.py
juliasaveliff: ok swagging will keep this in mind when cleaning up code

PR #81, opened 2020-01-31
location validation within file

benauriemma:


rackcity/views/it_instance_views.py
```diff
@@ -311,6 +314,17 @@ def instance_bulk_upload(request):
                     {"failure_message": failure_message},
                     status=HTTPStatus.BAD_REQUEST
                 )
+        # should this be whitespace/caps insenstive? Maybe later
+        if instance_data['hostname'] in hostnames_in_import:
```
juliasaveliff: will come back to this when we add that validation

PR #85, opened 2020-01-31
fields endpoints

juliasaveliff:
models returns:
```
{
    "fields": [
        "vendor",
        "model_number",
        "height",
        "display_color",
        "num_ethernet_ports",
        "num_power_ports",
        "cpu",
        "memory_gb",
        "storage",
        "comment"
    ]
}
```

instance fields:
```
{
    "fields": [
        "hostname",
        "model",
        "rack",
        "elevation",
        "owner",
        "comment"
    ]
}
```


rackcity/views/it_instance_views.py
```diff
@@ -459,3 +459,21 @@ def instance_page_count(request):
     instance_count = ITInstance.objects.all().count()
     page_count = math.ceil(instance_count / page_size)
     return JsonResponse({"page_count": page_count})
+
+
+@api_view(['GET'])
+def instance_fields(request):
```
benauriemma: Should we make this authenticated?

rackcity/views/it_model_views.py
```diff
@@ -381,6 +381,28 @@ def model_page_count(request):
     return JsonResponse({"page_count": page_count})
 
 
+@api_view(['GET'])
+def model_fields(request):
```
benauriemma: Same re: authentication

rackcity/views/it_instance_views.py
```diff
@@ -459,3 +459,21 @@ def instance_page_count(request):
...
+    """
+    return JsonResponse(
+        {"fields": [
+            'hostname',
+            'model',
+            'rack',
+            'elevation',
+            'owner',
+            'comment',
+        ]},
```
benauriemma: Is it hard to get the model or serializer to give you the fields dynamically?  I vaguely remember trying to do it on a model and you could but the response was ugly, but maybe serializer would be better since it contains a list of fields as a property?

I'm also fine with leaving this for now and refactoring later if we feel like it

clairecfu: Not sure if this is relevant, but I requested that we return the fields in a certain order (having elevation come right after rack) since the front end dynamically loads in these fields.

PR #86, opened 2020-02-01
Various bugs uncovered via FULL STACK TESTING!

juliasaveliff:


rackcity/models/it_instance.py
```diff
@@ -16,7 +16,7 @@ class ITInstance(models.Model):
         on_delete=models.CASCADE,
         verbose_name="related rack",
     )
-    owner = models.CharField(max_length=150)
+    owner = models.CharField(max_length=150, null=True, blank=True)
```
juliasaveliff: owner is OPTIONAL for instances

rackcity/views/it_instance_views.py
```diff
@@ -134,9 +134,9 @@ def instance_add(request):
     if not serializer.is_valid(raise_exception=False):
         failure_message += str(serializer.errors)
 
-    rack_id = data['rack']
-    elevation = data['elevation']
-    height = ITModel.objects.get(id=data['model']).height
+    rack_id = serializer.validated_data['rack'].id
+    elevation = serializer.validated_data['elevation']
+    height = serializer.validated_data['model'].height
```
juliasaveliff: Even though we know the values in data are valid since `serializer.is_valid` passed, if any integer values were passed as strings, they would still be string on `data['blah']` but would have been casted to ints on `validated_data['blah']`. This only matters here because we're using these 3 values for some logic so they need to be ints

rackcity/views/it_model_views.py
```diff
@@ -97,15 +97,16 @@ def model_modify(request):
 def model_height_change_valid(new_model_data, existing_model):
     if 'height' not in new_model_data:
         return True
-    if new_model_data['height'] <= existing_model.height:
+    new_model_height = int(new_model_data['height'])
```
juliasaveliff: Same issue as above. We don't use a serializer to validate the data on bulk upload responses because not all required fields are necessary. So this isn't automatically casted to an int yet

rackcity/views/it_instance_views.py
benauriemma: You

rackcity/models/it_instance.py
benauriemma: You

rackcity/views/it_instance_views.py
benauriemma: Amaze

rackcity/views/it_model_views.py
benauriemma: Me

PR #87, opened 2020-02-01
hostname RFC validation

juliasaveliff:


rackcity/views/it_instance_views.py
```diff
@@ -134,12 +134,13 @@ def instance_add(request):
...
-    height = serializer.validated_data['model'].height
+    if failure_message == "":
+        rack_id = serializer.validated_data['rack'].id
+        elevation = serializer.validated_data['elevation']
+        height = serializer.validated_data['model'].height
 
-    if is_location_full(rack_id, elevation, height):
-        failure_message += "Instance does not fit in this location. "
+        if is_location_full(rack_id, elevation, height):
+            failure_message += "Instance does not fit in this location. "
```
juliasaveliff: fix from previous bug PR. cannot be done if the serializer was invalid

rackcity/models/it_instance.py
```diff
@@ -1,10 +1,23 @@
+import re
+from django.core.exceptions import ValidationError
 from django.db import models
 from .it_model import ITModel
 from .rack import Rack
 
 
+def validate_hostname(value):
+    hostname_pattern = re.compile("[a-z][a-z0-9-]{0,61}[a-z0-9]")
```
juliasaveliff: thank you, patrick wang

rackcity/models/it_instance.py
```diff
@@ -22,3 +35,11 @@ class ITInstance(models.Model):
...
         ordering = ['hostname']
         verbose_name = 'instance'
+
+    def save(self, *args, **kwargs):
+        try:
+            validate_hostname(self.hostname)
+        except ValidationError as valid_error:
+            raise valid_error
+        else:
+            super(ITInstance, self).save(*args, **kwargs)
```
juliasaveliff: validators are NOT called on save automatically, unfortunately

rackcity/models/it_instance.py
benauriemma: Did you get this regex from somewhere or write it based on the rule?
1. This will require hostname to be at least two characters long -- is `a` not valid?
2. Can we document either on notion or in the comments the plain english version of what we think is a valid hostname

rackcity/models/it_instance.py
benauriemma: Could we override is_valid to call this instead of save? For bulk-upload this would make the difference between failing early and rejecting the whole file, vs starting to save and then failing halfway through. The req's mention something like "reject the file if there are naming issues or other conflicts" so I think we have to fail early there.

Alternatively we could call `validate_hostname` manually in bulk.

IF WE CHANGE THIS would need to make sure it works for instances/modify because iirc we don't call is_valid there, we just rely on save for the validation

rackcity/models/it_instance.py
```diff
@@ -22,3 +35,11 @@ class ITInstance(models.Model):
     class Meta:
         ordering = ['hostname']
         verbose_name = 'instance'
+
+    def save(self, *args, **kwargs):
+        try:
+            validate_hostname(self.hostname)
+        except ValidationError as valid_error:
+            raise valid_error
```
benauriemma: Catching the error and raising it is a lil redundant but I don't hate it for readability

rackcity/views/it_instance_views.py
benauriemma: I think I'm swayed to fail early.  Down to change it in the next PR

rackcity/views/it_instance_views.py
```diff
@@ -134,12 +134,13 @@ def instance_add(request):
     if not serializer.is_valid(raise_exception=False):
         failure_message += str(serializer.errors)
```
benauriemma: Wait does is_valid automatically call our new validator?? And then we just had to override save to make it also call the validator? If so that's big time

rackcity/models/it_instance.py
juliasaveliff: fixing regex, will add to documentation

rackcity/models/it_instance.py
juliasaveliff: feel that.

rackcity/views/it_instance_views.py
juliasaveliff: YES is_valid does indeed call all validators on the model because of how we defined our serializers. save needs to be overridden for the cases where save is called when is_valid is NOT, because save does not call all validators on the model (seems stoopid to me but that's what overriding is for)

rackcity/models/it_instance.py
```diff
@@ -1,10 +1,23 @@
+import re
+from django.core.exceptions import ValidationError
 from django.db import models
 from .it_model import ITModel
 from .rack import Rack
 
 
+def validate_hostname(value):
+    hostname_pattern = re.compile("[a-z]([a-z0-9-]{0,61}[a-z0-9])?")
```
juliasaveliff: added `( .. )?` around the last group so it needs 0 or more of the last shindig

benauriemma: 0 or *1 but the ? is correct

benauriemma: Buuuut I think capital letters are allowed sorry I didn't flag the first time

PR #96, opened 2020-02-03
Registering new users

clairecfu:


src/components/navigation/navigation.tsx
```diff
@@ -57,15 +57,24 @@ export class Navigation extends React.Component<NavigationPropsAll> {
                 <p></p>
               )}
             </NavbarGroup>
+
             <NavbarGroup align={Alignment.RIGHT}>
               {this.props.isAuthenticated ? (
```
brendancheng20: Should this be this.props.isAdmin or something instead, since only admins should be able to create new users?

src/components/login/register.tsx
```diff
@@ -1,42 +1,88 @@
-import { Button, Classes, InputGroup } from "@blueprintjs/core";
+import {
```
brendancheng20: Also similar question from my comment in navigation.tsx, should we make this route not even available to people if they aren't admins? i.e. what if they just enter this url directly?

PR #109, opened 2020-02-03
Distinct vendors bug fix

juliasaveliff:


benauriemma:
This is worse but I approve.

PR #118, opened 2020-02-04
Do not paginate if page not specified on /get-many

juliasaveliff:


rackcity/views/it_instance_views.py
```diff
@@ -40,20 +40,30 @@ def instance_list(request):
 
 @api_view(['POST'])
 @permission_classes([IsAuthenticated])
-def instance_page(request):
+def instance_many(request):
```
juliasaveliff: Renamed because not always returning a page of results anymore

PR #119, opened 2020-02-04
Import error handling and instructions

brendancheng20:
Implemented rudimentary error alerting (still need to use toasts but currently using the JS alert() function) and wrote bulk import instructions. Also fixed some formatting on report page


clairecfu:
we learned that alerts don't show up on safari, so we should try to avoid using these for error handling, so def track changing these to toasts/something else

PR #129, opened 2020-02-04
Hostname case insensitivity

juliasaveliff:
- Any instance added or modified NOT via bulk import must be case insensitive unique
   - ie if `server` is in the database you cannot add a new instance with hostname `Server` or modify a different instance to have hostname `Server`
- Any instance in BULK IMPORT will be detected as a MODIFY (not an add) if another case insensitive version of hostname already exists, and the hostname will be modified to have the new case explicitly specified in the import

### DIFF
--- a/rackcity/api/serializers/it_instance_serializers.py
+++ b/rackcity/api/serializers/it_instance_serializers.py
@@ -33,10 +33,6 @@ class RecursiveITInstanceSerializer(serializers.ModelSerializer):
     Recursively serializes all fields on ITInstance model, where model and
     rack fields are defined recursively (by all of their respective fields).
     """
-    hostname = serializers.CharField(validators=[
-        UniqueValidator(
-            queryset=ITInstance.objects.all(), lookup='iexact'
-        )])
     model = ITModelSerializer()
     rack = RackSerializer()
 
@@ -58,10 +54,6 @@ class BulkITInstanceSerializer(serializers.ModelSerializer):
     Serializes all fields on ITInstance model according to the format required
     for bulk export.
     """
-    hostname = serializers.CharField(validators=[
-        UniqueValidator(
-            queryset=ITInstance.objects.all(), lookup='iexact'
-        )])
     vendor = serializers.SlugRelatedField(
         source='model',
         slug_field='vendor',
--- a/rackcity/api/serializers/it_instance_serializers.py
+++ b/rackcity/api/serializers/it_instance_serializers.py
@@ -33,6 +33,10 @@ class RecursiveITInstanceSerializer(serializers.ModelSerializer):
     Recursively serializes all fields on ITInstance model, where model and
     rack fields are defined recursively (by all of their respective fields).
     """
+    hostname = serializers.CharField(validators=[
+        UniqueValidator(
+            queryset=ITInstance.objects.all(), lookup='iexact'
+        )])
     model = ITModelSerializer()
     rack = RackSerializer()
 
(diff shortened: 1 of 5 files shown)


rackcity/views/it_instance_views.py
```diff
@@ -368,24 +391,6 @@ def instance_bulk_upload(request):
...
-                return JsonResponse(
-                    {"failure_message": failure_message},
-                    status=HTTPStatus.BAD_REQUEST
-                )
-            potential_modifications.append(
-                {
-                    "existing_instance": existing_instance,
-                    "new_data": instance_data
-                }
-            )
```
juliasaveliff: moved above since this was `try...except...else` and now it's just `if...else...`

rackcity/views/rackcity_utils.py
```diff
@@ -165,7 +165,7 @@ def get_filter_arguments(data):
                 if filter_dict['match_type'] == 'exact':
                     filter_args.append(
                         {
-                            '{0}'.format(filter_field): filter_dict['value']
+                            '{0}__iexact'.format(filter_field): filter_dict['value']
```
juliasaveliff: fixes filter match issue

rackcity/api/serializers/it_instance_serializers.py
```diff
@@ -9,6 +10,11 @@ class ITInstanceSerializer(serializers.ModelSerializer):
     Serializes all fields on ITInstance model, where model and rack fields are
     defined by their pk only.
     """
+    hostname = serializers.CharField(validators=[
+        UniqueValidator(
+            queryset=ITInstance.objects.all(), lookup='iexact'
+        )])
+
```
juliasaveliff: case insensitive uniqueness will get validated whenever serializer.is_valid is called on this serializer

rackcity/views/it_instance_views.py
```diff
@@ -335,21 +335,44 @@ def instance_bulk_upload(request):
...
-            hostnames_in_import.add(instance_data['hostname'])
-        try:
-            existing_instance = ITInstance.objects.get(
-                hostname=instance_data['hostname'])
-        except ObjectDoesNotExist:
+            hostnames_in_import.add(instance_data_hostname_lower)
+
+        existing_instance_filtered = ITInstance.objects.filter(
+            hostname__iexact=instance_data['hostname'])
+        if len(existing_instance_filtered) == 1:
```
juliasaveliff: checks if any version of instance_data['hostname'] (i.e. any variation of `server`, `Server`, `seRvEr` etc) exists in the database

rackcity/views/it_instance_views.py
```diff
@@ -214,6 +214,13 @@ def instance_modify(request):
             value = ITModel.objects.get(id=data[field])
         elif field == 'rack':
             value = Rack.objects.get(id=data[field])
+        elif field == 'hostname':
+            if len(ITInstance.objects.filter(hostname__iexact=data[field])) != 0:
+                return JsonResponse(
+                    {"failure_message": "Instance with hostname '" +
+                        data[field].lower() + "' already exists."},
+                    status=HTTPStatus.BAD_REQUEST,
+                )
```
benauriemma: I'm confused by this part - if you're modifying an instance by changing the case of the hostname (I thought that was allowed) wouldn't this stop you from doing it?

juliasaveliff:
re-did all relevant test cases from yesterday, all working, merging

PR #130, opened 2020-02-04
registration fix

clairecfu:
- [x] Displays first and last name instead of displayname

### DIFF
--- a/src/forms/InstanceForm.tsx
+++ b/src/forms/InstanceForm.tsx
@@ -82,7 +82,7 @@ class InstanceForm extends React.Component<
 
     const { hostname, id, elevation, owner, comment } = instance;
     const model = instance.model ? instance.model.id : undefined;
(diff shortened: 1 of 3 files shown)


src/forms/InstanceForm.tsx
```diff
@@ -82,7 +82,7 @@ class InstanceForm extends React.Component<
 
     const { hostname, id, elevation, owner, comment } = instance;
     const model = instance.model ? instance.model.id : undefined;
-    const rack = instance.model ? instance.rack.id : undefined;
+    const rack = instance.rack ? instance.rack.id : undefined;
```
brendancheng20: Not a huge issue, but is there a reason that this change exists in this PR as well as in #128?

clairecfu: probably because of how i branched

PR #135, opened 2020-02-04
Validate owners and create endpoint

juliasaveliff:
Endpoint to get list of all usernames existing: 
- /api/usernames

When do validations on instance happen?
- on any instance serializer `.is_valid()`
- on `instance.save()`


rackcity/migrations/0012_auto_20200204_1812.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.0.2 on 2020-02-04 18:12
+
+from django.db import migrations, models
+import rackcity.models.it_instance
+
+
+class Migration(migrations.Migration):
```
juliasaveliff: I think I have two 0012 migrations again so we're going to have to merge these migrations when deploying but there are no conflicts so it's not a problem. Will try to do this on dev myself once both migration related PRs are merged to dev

clairecfu:
cool tracking this addition, lmk when this merges into dev

PR #139, opened 2020-02-04
Bulk export babyyyy

brendancheng20:


src/components/elementView/elementTable.tsx
```diff
@@ -64,6 +64,7 @@ interface IDragAndDrop {
 }
 
 interface IElementTableProps {
+  callback?: Function;
```
clairecfu: i think typescript convention for property functions includes specifying the params, param types, and return type, but if this works, that's fine for now

PR #141, opened 2020-02-04
implemented our own versions of a serializer field and a model field that handle empy strings better than the django models

benauriemma:


rackcity/api/serializers/fields/rackcity_serializer_fields.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+
+class RCIntegerField(IntegerField):
+
+    def __init__(self, **kwargs):
+        self.max_value = kwargs.pop('max_value', None)
+        self.min_value = kwargs.pop('min_value', None)
+        super().__init__(**kwargs)
+        if self.max_value is not None and self.max_value != "":
```
benauriemma: Had to override this method because the "" broke comparison to min and max values. Only change is from `if self.max_value is not None:` to `if self.max_value is not None and self.max_value != "":`

rackcity/api/serializers/fields/rackcity_serializer_fields.py
```diff
@@ -0,0 +1,35 @@
...
+            message = lazy_format(
+                self.error_messages['max_value'], max_value=self.max_value)
+            self.validators.append(
+                MaxValueValidator(self.max_value, message=message))
+        if self.min_value is not None and self.min_value != "":
```
benauriemma: Same here

rackcity/api/serializers/fields/rackcity_serializer_fields.py
```diff
@@ -0,0 +1,35 @@
...
+            message = lazy_format(
+                self.error_messages['min_value'], min_value=self.min_value)
+            self.validators.append(
+                MinValueValidator(self.min_value, message=message))
+
+    def to_internal_value(self, data):
+        if isinstance(data, str) and len(data) > self.MAX_STRING_LENGTH:
+            self.fail('max_string_length')
+        if isinstance(data, str) and data == "":
+            return None
```
benauriemma: These two lines are the reason we implement our own serializer IntegerField -- the Django one would carry on to the try statement and fail there because "" isn't a number

rackcity/api/serializers/it_model_serializers.py
```diff
@@ -1,8 +1,19 @@
...
 class ITModelSerializer(serializers.ModelSerializer):
+
+    # This is the field for the serializers
+    # It overrides appropriate methods to work with empty strings when you call validate
+    memory_gb = RCIntegerField(
+        allow_null=True,
+        max_value=2147483647,
+        min_value=0,
+        required=False
+    )
```
benauriemma: Because we're using a `serializers.ModelSerializer`, it would normally automatically detect that the field on the ITModel model is a `models.PositiveIntegerField`, and the serializer should use a `serializer.IntegerField` and set the values that you see here.  Instead, we force it to use our own version that overrides some methods of `serializer.IntegerField`

rackcity/models/fields/rackcity_model_fields.py
```diff
@@ -0,0 +1,8 @@
+from django.db.models import PositiveIntegerField
+
+
+class RCPositiveIntegerField(PositiveIntegerField):
+    def get_prep_value(self, value):
+        if value == '':
+            return None
```
benauriemma: This is the big change on the model serializer.  Before saving to the database, if the data is "", this will force it to None.  This is relevant when we didn't validate the serializer (like on modify) or (I believe) if we validated the serializer but used the un-validated data anyway

rackcity/views/it_model_views.py
```diff
@@ -296,7 +296,7 @@ def model_bulk_upload(request):
         model_serializer = ITModelSerializer(data=model_data)
         if not model_serializer.is_valid():
             failure_message = str(model_serializer.errors)
-            failure_message = "At least one provided model was not valid. "+failure_message
+            failure_message = "At least one provided model was not valid! "+failure_message
```
benauriemma: Accident but I like the flair

rackcity/models/fields/rackcity_model_fields.py
juliasaveliff: cool

juliasaveliff:
migrations look good 2 me

PR #143, opened 2020-02-04
specific error messages for racks

juliasaveliff:
examples below

deleting nonexistent racks:
```
"The range of racks L-M 1-2 cannot be deleted. The following racks within this range do not exist: L1, L2, M1, M2. "
```

deleting nonempty racks: 
```
"The range of racks A-B 1-2 cannot be deleted. The following racks within this range contain instances: A1, A2, B1, B2. "
```

deleting nonexistent and nonempty racks: 
```
"The range of racks A-A 1-5 cannot be deleted. The following racks within this range do not exist: A3, A4, A5. The following racks within this range contain instances: A1, A2. "
```

### DIFF
--- a/rackcity/api/objects/rack_range.py
+++ b/rackcity/api/objects/rack_range.py
@@ -72,6 +72,10 @@ class RackRangeSerializer(serializers.Serializer):
                 )
             ]
 
+    def get_row_range_as_string(self):
+        row_range = self.get_row_range()
+        return row_range[0] + "-" + row_range[1]
+
     def get_number_range(self):
         """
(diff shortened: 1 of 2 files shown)


juliasaveliff:
@benauriemma changed the logic in delete but it works the same way / I like it better now

juliasaveliff:
it can get a little hefty on the front end, but i think it's worth being explicit according to all the piazza clarifications 

image

juliasaveliff:
update: I am switching order of errors because I think containing instances has priority over not existing, but that's not a biggie

PR #145, opened 2020-02-04
fixed bug where it thought null and empty string werent the same for string fields in bulk upload

benauriemma:


rackcity/views/rackcity_utils.py
```diff
@@ -83,13 +84,32 @@ def records_are_identical(existing_data, new_data):
...
                 return False
     return True
 
 
+def int_string_comparison(existing_value, new_value):
+    return (
+        isinstance(existing_value, int)
+        and isinstance(new_value, str)
+        and int(new_value) == existing_value
+    )
```
juliasaveliff: this will throw a ValueError exception if new_value cannot be cast as an int (i.e. is "hi"). Is that the intended behavior?

benauriemma: Just tested and this actually fails earlier with a good message because the validator would throw the error

juliasaveliff: I love that for us

PR #158, opened 2020-02-10
Front end cleanup round 1

clairecfu:
Renaming classes/files to be more consistent
Moving files into different folders


clairecfu:
@brendancheng20 I think this should merge into dev with no conflicts, since I already handled all the issues with the file renaming stuff in this branch/there should not be duplicates

PR #160, opened 2020-02-12
Filtering paging/column names 

clairecfu:
- fixed paging bug on filtering
- show table column names even if no data present
- add getFieldNames() util function 
- remove spinner
- column widths

### DIFF
--- a/src/components/elementView/elementTable.tsx
+++ b/src/components/elementView/elementTable.tsx
@@ -1,6 +1,5 @@
 import {
   Icon,
-  Spinner,
   HTMLSelect,
   Position,
   IToastProps,
@@ -18,8 +17,7 @@ import {
   ElementType,
   isModelObject,
   isRackObject,
-  RackRangeFields,
-  getHeaders
+  RackRangeFields
 } from "../../utils/utils";
 import DragDropList from "./dragDropList";
 import "./elementView.scss";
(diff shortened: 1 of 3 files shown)


src/components/elementView/elementTable.tsx
```diff
@@ -451,12 +457,14 @@ class ElementTable extends React.Component<
...
     });
     this.updateFilterData(items);
   };
-  getFieldNames = () => {
+  setFieldNames = () => {
     let fields: Array<string> = [];
+
     if (this.state.items && this.state.items.length > 0) {
       Object.keys(this.state.items[0]).forEach((col: string) => {
         if (col === "model") {
```
brendancheng20: Is this the function you were referring to that dynamically renders column names? And if so, why do you only push if col === model?

clairecfu: oh, I think the rest of the method was just hidden,I have a special case for model because i need to break it into model__vendor and model__model_number fields. i account for the other columns here :

 ```       fields.push("model__vendor");
          fields.push("model__model_number");
        } else if (col !== "id") {
          fields.push(col);
        }
      });```

clairecfu: Also, no this was not the function I was referring to, although this one probably also works for detailview. I added one in utils.tsx that actually calls the backend endpoint that returns the fields. This one gets the fields from the current data. Either one would work.

PR #161, opened 2020-02-12
adding network ports to it_model

clairecfu:
- [x] Update ITModel
- [x] Update API Object with new model
- [x] Update ITModelSerializer
- [x] Allow network ports to be specified with a count or a list (create/modify endpoint)


clairecfu:
update: tested create/modify/get endpoints and verified that they work via postman

rackcity/api/serializers/it_model_serializers.py
```diff
@@ -46,13 +40,7 @@ class BulkITModelSerializer(serializers.ModelSerializer):
     Serializes all fields on ITModel model according to the format required
     for bulk export.
     """
-    ethernet_ports = RCIntegerField(
-        source='num_ethernet_ports',
-        allow_null=True,
-        max_value=2147483647,
-        min_value=0,
-        required=False
-    )
```
juliasaveliff: I would recommend keeping this serializer unchanged since we're not sure how this is going to affect the implementation of model bulk imports yet. I'd say it's likely that we're going to get rid of this field on the serializer, but I think it'd be better to keep this for now and only take it out once we're sure

rackcity/api/serializers/it_model_serializers.py
```diff
@@ -75,7 +63,7 @@ class Meta:
             'model_number',
             'height',
             'display_color',
-            'ethernet_ports',
+            'network_ports',
```
juliasaveliff: switch this back too ^

rackcity/models/it_model.py
```diff
@@ -19,7 +20,13 @@ class ITModel(models.Model):
         default='#394B59',
         validators=[validate_display_color],
     )
-    num_ethernet_ports = RCPositiveIntegerField(null=True, blank=True)
+
+    network_ports = ArrayField(
+        models.CharField(max_length=150, blank=True),
```
juliasaveliff: per our discussion, I'd suggest we do not set `blank=True` here because we would actually set the "default" values ("1", "2", etc) a step further away from the model. we wouldn't want a row in the table to contain an array of empty strings since these would not be valid network port names

juliasaveliff: this is the only change that requires a migration BTW

PR #163, opened 2020-02-12
Bulk Import UI Refactor

brendancheng20:
- Moved bulk import button from navbar to elementTable, per class discussion
- Cleaned up bulk import component by moving instructions to a separate, static importInstructions.tsx component


src/components/elementView/elementTab.tsx
```diff
@@ -161,14 +163,26 @@ class ElementTab extends React.Component<ElementViewProps, ElementViewState> {
...
+            console.log(this.state.filters)
           }}
         />
+        {this.props.isAdmin ? (
+          <div>
+            <AnchorButton
+              onClick={() => this.props.history.push("/bulk-upload")}
+              className="add"
+              icon="export"
+              text="Upload from CSV file"
```
clairecfu: small nit on the wording of the button. Since it is now next to `Add Instance` button, maybe this could say `Add from File`? and make these buttons "minimal" style so they match the `add instance` button.

clairecfu:
Only comment is about the naming of the button, can you address that before merging?

PR #166, opened 2020-02-13
Create custom registration endpoint

juliasaveliff:
Front end changes
- Call new endpoint /api/users/add instead of rest-auth registration endpoint
- Pass firstName and lastName to request
- Grab these values from their fields

Back end changes
- New endpoint /api/users/add wrapper for rest-auth registration, update with admin-only permissions
- Custom serializer to allow first name and last name input


djreact/urls.py
```diff
@@ -5,7 +5,6 @@
 urlpatterns = [
     path('api-auth/', include('rest_framework.urls')),
     path('rest-auth/', include('rest_auth.urls')),
-    path('rest-auth/registration/', include('rest_auth.registration.urls')),
```
juliasaveliff: No longer calling djangorestauth registration endpoint directly so we shouldn't expose this URL directly in the API. Must go through new wrapper view

rackcity/api/serializers/user_serializers.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class RegisterNameSerializer(RegisterSerializer):
+    first_name = serializers.CharField(required=False)
+    last_name = serializers.CharField(required=False)
+
+    def custom_signup(self, request, user):
+        user.first_name = self.validated_data.get('first_name', '')
+        user.last_name = self.validated_data.get('last_name', '')
+        user.save(update_fields=['first_name', 'last_name'])
```
juliasaveliff: this method is provided by djangorestauth to add additional fields to registration if wanted. it does not override default signup (so username, email, passwords still get updated too)

rackcity/views/user_views.py
```diff
@@ -2,7 +2,14 @@
...
 from rest_framework.decorators import permission_classes, api_view
-from rest_framework.permissions import IsAuthenticated
+from rest_framework.permissions import IsAuthenticated, IsAdminUser
+from rest_auth.registration.views import RegisterView
+from rackcity.api.serializers import RegisterNameSerializer
+
+
+class RegisterNameView(RegisterView):
+    serializer_class = RegisterNameSerializer
+    permission_classes = [IsAdminUser]
```
juliasaveliff: implementing this as a class-based view (not function-based view) so we can extend the existing RegisterView provided by djangorestauth

src/forms/auth/register.tsx
```diff
@@ -20,8 +20,8 @@ interface RegistrationFormProps {
   error: string;
   token: string;
 }
-var console:any = {};
-console.log = function(){};
+var console: any = {};
+console.log = function () { };
```
juliasaveliff: linting

PR #167, opened 2020-02-13
Create endpoint to get logged in user info

juliasaveliff:


rackcity/views/it_model_views.py
```diff
@@ -496,7 +496,7 @@ def model_fields(request):
 @api_view(['GET'])
 def i_am_admin(request):
     print("yeah")
-    if(request.user.is_superuser):
+    if(request.user.is_staff):
```
juliasaveliff: technically staff == admin in Django world. superuser is an all powerful version of admin that can never be defeated. if you are superuser, you are also staff. if you are staff, you are not necessarily superuser

clairecfu: is staff the new backend representation of admin? Do we also need to be able to distinguish between the "admin admin" and other admin users?

juliasaveliff: see above comment!! staff & superuser are Django definitions

PR #175, opened 2020-02-14
Created get-all datacenters endpoint

brendancheng20:
- [ ] works on postman, but definitely worth checking out on your own branch to be sure it doesn't break anything for you


rackcity/views/datacenter_views.py
```diff
@@ -0,0 +1,32 @@
...
+from rackcity.models import Datacenter
+from django.core.exceptions import ObjectDoesNotExist
+from rackcity.api.serializers import DatacenterSerializer
+from rest_framework.decorators import permission_classes, api_view
+from rest_framework.permissions import IsAuthenticated, IsAdminUser
+from rest_framework.parsers import JSONParser
+from rest_framework.pagination import PageNumberPagination
+from http import HTTPStatus
+import math
+from io import StringIO
```
brendancheng20: Are there gonna be linting problems here? I can remove if need be but idk how backend stuff works

rackcity/views/datacenter_views.py
```diff
@@ -0,0 +1,32 @@
...
+        Return List of all datacenters.
+    """
+    try:
+        datacenters = Datacenter.objects.all()
+        serializer = DatacenterSerializer(datacenters, many=True)
+        return JsonResponse(
+            {"datacenters": serializer.data},
+            status=HTTPStatus.OK
+        )
+    except Datacenter.Error:
```
brendancheng20: Wasn't sure what error to catch so I just threw a standard error for now

rackcity/views/datacenter_views.py
benauriemma: It should underline the ones that are unused - I'd delete all the ones you don't need

rackcity/views/datacenter_views.py
benauriemma: I wouldn't do this because it can lead to covering up an error with an incorrect failure message which can be super misleading. I'd say either
- figure out what the error type is when there are no datacenters (might be ObjectDoesNotExist or something like that) and catch that specifically
- don't catch it at all and just let the server 500. We don't want that but if it's an error we didn't know we could get then it should fail
- catch any error by saying `except Exception as error` and then make the failure message `str(error)`

benauriemma: Also I think best practices is to have only the line you think might fail in the try: and then do the rest either in an else: or outside of the try/except block

rackcity/views/datacenter_views.py
```diff
@@ -0,0 +1,32 @@
...
+        failure_message = "No datacenters"
+        return JsonResponse(
+            {"failure_message": failure_message},
+            status=HTTPStatus.BAD_REQUEST,
```
benauriemma: This is picky but if there are no datacenters this isn't a bad request - I'd use a different status here, I think there might be one about a resource not existing or something like that

benauriemma: Actually now that I'm thinking about it I'm not sure this should fail at all.  I think maybe it should return 200 and just make the datacenters list empty

brendancheng20: will do

PR #176, opened 2020-02-14
added datacenter to rack create, get, and delete

benauriemma:
not on summary or get-all as those might get depecated soon

### DIFF
--- a/rackcity/api/objects/rack_range.py
+++ b/rackcity/api/objects/rack_range.py
@@ -1,8 +1,13 @@
 from rest_framework import serializers
 from rackcity.models.rack import validate_row_letter
+from rackcity.models import Datacenter
 
 
 class RackRangeSerializer(serializers.Serializer):
+    datacenter = serializers.PrimaryKeyRelatedField(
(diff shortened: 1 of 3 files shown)


rackcity/api/objects/rack_range.py
```diff
@@ -1,8 +1,13 @@
 from rest_framework import serializers
 from rackcity.models.rack import validate_row_letter
+from rackcity.models import Datacenter
 
 
 class RackRangeSerializer(serializers.Serializer):
+    datacenter = serializers.PrimaryKeyRelatedField(
+        label='datacenter',
+        queryset=Datacenter.objects.all()
+    )
```
juliasaveliff: if you're going to add this to the rack range serializer, i think it should be renamed to something like rack filter serializer, because datacenter is not part of a rack range

juliasaveliff: if you're going to add this to the rack range serializer, i think it should be renamed to something like rack filter serializer, because datacenter is not part of a rack range

PR #182, opened 2020-02-15
instance? never heard of her

benauriemma:


benauriemma:
Do not merge label is because I ran make migrations but did not run migrate - I'll do that when we're ready to merge but it may cause trouble with existing data

benauriemma:
Ready to go - got a little creative on the migration but everything should be intact

PR #185, opened 2020-02-15
Log model/serializer/endpoint

juliasaveliff:
example response from paginated /get-many endpoint

```
{
    "logs": [
        {
            "id": 4,
            "date": "2020-02-17T04:32:03.269754Z",
            "log_content": "this is a test log with all fields populated",
            "user": 1,
            "related_assets": [
                134,
                130
            ],
            "related_model": 35
        },
        {
            "id": 3,
            "date": "2020-02-15T21:04:06.271660Z",
            "log_content": "this is also a test log",
            "user": 16,
            "related_assets": [],
            "related_model": null
        },
        {
            "id": 2,
            "date": "2020-02-15T21:03:44.031441Z",
            "log_content": "this is a test log",
            "user": 1,
            "related_assets": [],
            "related_model": null
        }
}
```


rackcity/admin.py
```diff
@@ -1,7 +1,8 @@
 from django.contrib import admin
 
-from .models import Asset, ITModel, Rack
+from .models import Asset, Log, ITModel, Rack
 
 admin.site.register(Asset)
+admin.site.register(Log)
```
benauriemma: What is this and should I have added models that I created here 😳

rackcity/api/serializers/log_serializers.py
```diff
@@ -0,0 +1,15 @@
...
+class LogSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Log
+        fields = (
+            'id',
+            'date',
+            'log_content',
+            'user',
+            'related_assets',
+            'related_model',
```
benauriemma: Is it model singular and assets plural because network connections can be between assets but everything else would only be one resource?

benauriemma: Would you not log actions on racks or datacenters?

benauriemma: Also I trust you have a plan for how you're displaying the logs and how you're linking to models/asset and how that all relates to how you serialize this

rackcity/migrations/0008_merge_20200217_0437.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.0.2 on 2020-02-17 04:37
```
benauriemma: Great migrations

rackcity/models/log.py
```diff
@@ -0,0 +1,30 @@
...
+from django.contrib.auth.models import User
+from django.db import models
+from .asset import Asset
+
+
+class Log(models.Model):
+    date = models.DateTimeField(
+        auto_now=True,
+        blank=True,
+        editable=False,
```
benauriemma: `editable=False` 😗👌🏼

rackcity/models/log.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+    log_content = models.TextField()
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
```
benauriemma: I don't think you want `on_delete=models.CASCADE,` here - that means if the user, model, or asset a log references is deleted, all logs about it will be deleted. You probably want either `DO_NOTHING` or `SET_NULL`

benauriemma: Wondering if we use this correctly everywhere else 🤔 there's also a `PROTECT ` option which won't let you delete the referenced object until you delete the objects that reference it.  This could have been nice on the model and rack fields of Asset

rackcity/models/log.py
```diff
@@ -0,0 +1,30 @@
...
+        verbose_name="related user",
+    )
+    related_assets = models.ManyToManyField(
+        Asset,
+        verbose_name="related assets",
+        blank=True,
+    )
+    related_model = models.ForeignKey(
+        ITModel,
+        on_delete=models.CASCADE,
```
benauriemma: Ditto

rackcity/models/log.py
```diff
@@ -0,0 +1,30 @@
...
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        verbose_name="related user",
+    )
+    related_assets = models.ManyToManyField(
+        Asset,
+        verbose_name="related assets",
+        blank=True,
+    )
```
benauriemma: Will on_delete default to something here? Or is that not a thing on many to many fields

rackcity/admin.py
juliasaveliff: it just lets you look at the table from django `/admin/` endpoint, just wanted it while testing not super impactful

rackcity/api/serializers/log_serializers.py
juliasaveliff: 1. yes, network connections are the ONLY reason there would ever be more than one asset linked. model doesn't have anything like that. 
2. yes logging actions on these, but the related field will be used for linking, and we do NOT need to link racks/datacenters/users because the link should bring you to the detail view, and only assets and models have detail views
3. TRUST FALL

rackcity/migrations/0008_merge_20200217_0437.py
juliasaveliff: "Good vibrations -- I mean migrations" - beach boys

rackcity/models/log.py
juliasaveliff: my GOD I did not actually know that's what CASCADE meant!!!!!! changing

rackcity/models/log.py
juliasaveliff: many to many does not have on_delete. upon investigation i have come to learn that many to many delete behavior is similar to cascade. once i actually start creating these logs i might come back and edit this field.......tbd

rackcity/admin.py
benauriemma: That looks helpful actually I've never used it

rackcity/models/log.py
benauriemma: If you're concerned about this you could have a `related_asset` and `other_related_asset`  field each as a foreign key... haha jk... unless?

rackcity/models/log.py
benauriemma: I did not either but stumbled upon that this morning with the port tables. Crazy stuff

rackcity/models/log.py
juliasaveliff: I tried this too LOL and you can't have multiple foreign keys to the same model that look the same


haha unless.

PR #188, opened 2020-02-16
Updated rack model for pdu management

brendancheng20:
- Created is_powered boolean field

- Developed logic in rack save() function to set is_powered value by automatically checking if the rack is 1) in the "rtp1" datacenter, 2) has row letters A-E, and 3) has rack number 1-19, as those are the only racks that can be powered by the pdu software right now


rackcity/models/rack.py
```diff
@@ -31,7 +32,17 @@ class Meta:
                 name='unique rack letter and number per datacenter'),
         ]
 
+    def char_range(self, c1, c2):
+        """Generates the characters from `c1` to `c2`, inclusive."""
+        for c in range(ord(c1), ord(c2)+1):
+            yield chr(c)
```
benauriemma: Is it possible to make this not take self as an argument so you can just call `char_range()` instead of `self.char_range()`?

rackcity/models/rack.py
```diff
@@ -22,6 +22,7 @@ class Rack(models.Model):
     )
     rack_num = models.PositiveIntegerField()
     height = models.PositiveIntegerField(default=42)
+    is_powered = models.BooleanField(default=False)
```
benauriemma: This is a little picky but I don't think `is_powered` is an accurate name since all racks have PDUs - I think something like `has_networked_pdu` or `is_networx_pro` or `is_network_controlled` would be better

rackcity/models/rack.py
brendancheng20: It would not let me unfortunately

rackcity/models/rack.py
brendancheng20: Sounds good, I'll update and push again

PR #190, opened 2020-02-16
assets, network ports, power ports, and pdu ports - hopefully done making our database a n g r y

benauriemma:


rackcity/models/rack.py
```diff
@@ -49,3 +49,16 @@ def save(self, *args, **kwargs):
...
+        from rackcity.models import PDUPort
+        if len(PDUPort.objects.filter(rack=self.id)) == 0:
+            for left_right in ["L", "R"]:
+                for port_number in range(1, 25):  # 1 through 24 inclusive
+                    pdu_port = PDUPort(
+                        rack=self,
+                        left_right=left_right,
+                        port_number=port_number,
+                    )
+                    pdu_port.save()
```
juliasaveliff: I think we need similar code on rack.delete() so that the rack's associated rows in the pduport table are also deleted?

rackcity/models/asset.py
```diff
@@ -77,3 +78,28 @@ def save(self, *args, **kwargs):
                         self.asset_number = asset_number
                         break
             super(Asset, self).save(*args, **kwargs)
+            self.add_network_ports()
+            self.add_power_ports()
```
juliasaveliff: same as for racks....I think we need similar code on asset.delete() so that the asset's associated rows in the networks ports & power ports tables are also deleted?

rackcity/models/pdu_port.py
```diff
@@ -0,0 +1,32 @@
...
+from django.core.validators import MaxValueValidator, MinValueValidator
+
+
+class PDUPort(models.Model):
+    rack = models.ForeignKey(
+        Rack,
+        on_delete=models.CASCADE,
+        verbose_name="rack",
+    )
+    LEFT_RIGHT_CHOICES = [
```
brendancheng20: What is this for?

rackcity/models/asset.py
```diff
@@ -5,6 +5,7 @@
 from django.db import models
 from .it_model import ITModel
 from .rack import Rack
+#from .network_port import NetworkPort
```
juliasaveliff: commented import 🤪

rackcity/models/asset.py
```diff
@@ -77,3 +78,28 @@ def save(self, *args, **kwargs):
...
+
+    def add_network_ports(self):
+        from rackcity.models import NetworkPort
+        if len(NetworkPort.objects.filter(asset=self.id)) == 0:  # only add ports for new assets
+            model = self.model
+            for network_port_name in model.network_ports:
+                network_port = NetworkPort(
+                    asset=self,
+                    port_name=network_port_name,
+                )
```
juliasaveliff: where are we going to handle the writing of MAC addresses to the NetworkPort table?

rackcity/models/asset.py
```diff
@@ -77,3 +78,28 @@ def save(self, *args, **kwargs):
...
+                network_port.save()
+
+    def add_power_ports(self):
+        from rackcity.models import PowerPort
```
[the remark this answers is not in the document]
benauriemma: Was getting funky stuff with circular imports because Asset imports *Port and *Port imports Asset.  Looked up solutions and the only one that wasn't crazy is importing it in the method.  I think I could also call it by its fully qualified name instead of importing but not sure that's any better

rackcity/models/asset.py
```diff
@@ -77,3 +78,28 @@ def save(self, *args, **kwargs):
                         self.asset_number = asset_number
                         break
             super(Asset, self).save(*args, **kwargs)
+            self.add_network_ports()
+            self.add_power_ports()
+
+    def add_network_ports(self):
+        from rackcity.models import NetworkPort
```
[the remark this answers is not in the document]
benauriemma: See above

rackcity/models/network_port.py
```diff
@@ -0,0 +1,34 @@
...
+        max_length=150
+        # electing not to validate this as it is not user entered
+    )
+    mac_address = models.CharField(
+        max_length=17,
+        unique=True,
+        null=True,
+        blank=True,
+        # force this to lowercase and make delimeters :
+    )
```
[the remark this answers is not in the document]
benauriemma: Yep - I haven't implemented the setting of MAC addresses yet, but have a task for that

rackcity/models/network_port.py
```diff
@@ -0,0 +1,34 @@
...
+        null=True,
+        blank=True,
+    )
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=['asset', 'port_name'],
+                name='unique network port names on assets'),
+        ]
```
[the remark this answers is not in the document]
benauriemma: Whoops I assumed that was implemented but def should be

rackcity/models/rack.py
benauriemma: NICE CATCH I almost forgot this and then was proud of myself for realizing HOWEVER I tested and when you delete the rack all of the pdu ports get deleted as well which is the behavior we want.  Nice thing that we get for free from django/postgres given our new database strats

rackcity/models/asset.py
benauriemma: See above

rackcity/models/asset.py
benauriemma: Nooo I was doing so good. Will change

rackcity/models/pdu_port.py
benauriemma: The choices field in Django makes you give is a tuple of what's stored in the database and what you display (think it's a legacy thing) - in our case both are the same

rackcity/models/asset.py
benauriemma: Was planning to do the same place where I handle writing the connections, which would be in the function views after the asset has been saved

rackcity/models/asset.py
benauriemma: Fixed

rackcity/models/asset.py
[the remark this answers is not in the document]
juliasaveliff: stars, they're just like us - everyone has some circular dependencies in their closet

rackcity/models/rack.py
juliasaveliff: that's crazy. how beautiful

rackcity/models/asset.py
juliasaveliff: swaggy

PR #192, opened 2020-02-16
User table controls 

clairecfu:
- [x] adding inline add admin/remove admin buttons to user table 
- [x] modified users endpoint to return admin status, and updated front end to reflect this new field 
- [x] general elementTab/elementTable refactoring


rackcity/api/serializers/user_serializers.py
```diff
@@ -22,4 +22,5 @@ class Meta:
             'email',
             'first_name',
             'last_name',
+            'is_staff'
```
clairecfu: adding this to the serializer -> in the future should we return something that is more user friendly?  ex. `type: superadmin|admin|local user|remote user|`

src/components/elementView/elementTable.tsx
```diff
@@ -611,6 +613,33 @@ class ElementTable extends React.Component<
...
+      console.log("NOT AN ADMIN");
+      return (
+        <AnchorButton
+          className="button-table"
+          intent="primary"
+          icon="user"
+          minimal
+          text="Remove Admin "
+        />
+      );
```
clairecfu: new in line buttons

clairecfu:
rest of the fe changes are me playing around with css/styling changes

PR #193, opened 2020-02-17
Datacenter UI 

clairecfu:
- [x] shared current datacenter value between Racks and Assets Tab
- [x] styling for a smoother UI experience
- [x] handling logic for racks if ALL datacenters are selected

### DIFF
--- a/src/components/elementView/rackSelectView.tsx
+++ b/src/components/elementView/rackSelectView.tsx
@@ -1,4 +1,11 @@
-import { Button, Callout, Intent, Switch, FormGroup, MenuItem } from "@blueprintjs/core";
+import {
+  Button,
+  Callout,
+  Intent,
+  Switch,
+  FormGroup,
+  MenuItem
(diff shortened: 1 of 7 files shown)


src/components/elementView/rackSelectView.tsx
```diff
@@ -69,39 +63,10 @@ class RackSelectView extends React.Component<
     const headers = getHeaders(this.props.token);
     this.props.submitForm(this.state.values, headers);
   };
-  renderRackOptions(range: boolean) { }
-  componentDidMount() { }
```
brendancheng20: Just to be sure, was this whole section removed because we no longer need to get Datacenters, as they are part of the shared state even for racks?

clairecfu: rackselectview basically only handles allowing users to input a rack range, so the datacenter that the rack range is for is handled at a higher level component

PR #194, opened 2020-02-17
Finished up datacenter epic

brendancheng20:
- Modify, add, and delete endpoints on back end

- Front end support for all datacenter operations

- Landing page elementTable now displays all datacenters and gives the user the option to modify add or delete


src/components/elementView/rackTab.tsx
```diff
@@ -142,7 +142,7 @@ class RackTab extends React.Component<RackTabProps, RackTabState> {
       loading: true
     });
     axios
-      .get(API_ROOT + "api/racks/get-all", getHeaders(this.props.token))
+      .post(API_ROOT + "api/racks/get-many", { sort_by: [], filters: [] }, getHeaders(this.props.token))
```
clairecfu: what endpoint are you calling? i thought we only had get-all or get, but did something get added but not updated in the endpoints table?

src/components/elementView/elementTabContainer.tsx
```diff
@@ -40,7 +40,12 @@ class ElementTabContainer extends React.Component<
         />
 
         <Tab className="tab" id="rack" title="Racks" panel={<RackTab />} />
-        <Tab className="tab" id="datacenter" title="Datacenters" />
+        <Tab
+          className="tab"
+          id="datacenter"
+          title="Datacenters"
+          panel={<ElementTab {...this.props} element={ElementType.DATACENTER} />}
+        />
```
clairecfu: we should hide this from the user if the user is not admin, so there should be some logic for that here

PR #197, opened 2020-02-17
mac addresses at api layer

benauriemma:
The meat of this is adding mac addresses to assets/add, assets/modify, and assets/<id>. I also changed some references of hostname to be asset number in a few places, and fixed a couple bugs that came up along the way


rackcity/views/asset_views.py
```diff
@@ -166,18 +182,58 @@ def asset_add(request):  # need to make network and power connections here
             )
     if failure_message == "":
         try:
-            serializer.save()
-            return HttpResponse(status=HTTPStatus.CREATED)
+            asset = serializer.save()
```
benauriemma: serializer.save() returns the object, which we use to get the id

rackcity/views/asset_views.py
```diff
@@ -166,18 +182,58 @@ def asset_add(request):  # need to make network and power connections here
...
-
+        else:
+            try:
+                save_mac_addresses(
+                    asset_data=data,
+                    asset_id=asset.id
+                )
+            except MacAddressException as error:
+                failure_message += "Some mac addresses couldn't be saved. " + \
+                    str(error)
```
benauriemma: Only the invalid ones aren't saved, and the (user friendly) errors are listed here

rackcity/views/asset_views.py
```diff
@@ -230,19 +286,47 @@ def asset_modify(request):
...
+                    status=HTTPStatus.BAD_REQUEST,
+                )
+            value = data[field]
         else:
             value = data[field]
         setattr(existing_asset, field, value)
 
     try:
         existing_asset.save()
+        # will need to save mac addresses and connections here
```
benauriemma: Still need to do connections here in next PR

rackcity/views/asset_views.py
```diff
@@ -420,6 +506,7 @@ def asset_bulk_upload(request):  # need to make network and power connections he
     for asset_to_add in assets_to_add:
         records_added += 1
         asset_to_add.save()
+        # will need to save mac addresses and connections here
```
benauriemma: Didn't address bulk stuff in this PR because we'll get mac addresses in a very different format from bulk

rackcity/views/asset_views.py
```diff
@@ -374,8 +458,9 @@ def asset_bulk_upload(request):  # need to make network and power connections he
                 validate_location_modification(
                     asset_data, existing_asset)
             except Exception:
+                # this could be problematic if neither hostame or asset number isn't there
```
benauriemma: Just a flag for when I come back for bulk stuff

rackcity/views/asset_views.py
```diff
@@ -135,7 +136,22 @@ def asset_detail(request, id):
...
+def get_mac_addresses(asset_id):
+    try:
+        ports = NetworkPort.objects.filter(asset=asset_id)
+    except ObjectDoesNotExist:
+        return
+    mac_addresses = {}
+    for port in ports:
+        if port.mac_address:
+            mac_addresses[port.port_name] = port.mac_address
+    return mac_addresses
```
juliasaveliff: just curious - why put this step here instead of adding it onto the RecursiveAssetSerializer?

rackcity/views/asset_views.py
juliasaveliff: sounds good

rackcity/views/asset_views.py
juliasaveliff: 💪

PR #202, opened 2020-02-18
support for multiline comments

clairecfu:
- allow input asset/model form
- show new lines on detailview


src/components/elementView/detailedView/propertiesView.scss
```diff
@@ -1,5 +1,6 @@
 .propsview .row-props {
   display: flex;
+  white-space: pre-line;
```
clairecfu: important change!! had to add this so that propertiewView renders new lines. If you do a refactor of propertiesview,  this should def still be included so we can see new lines rendered correctly

PR #203, opened 2020-02-18
Revoke and grant admin endpoints

juliasaveliff:


rackcity/views/user_views.py
```diff
@@ -213,6 +213,131 @@ def who_am_i(request):
...
 
+@api_view(['GET'])
+def i_am_admin(request):
+    """
+    Returns whether logged in user is an admin user.
+    """
+    if(request.user.is_staff):
+        return JsonResponse({"is_admin": True})
+    else:
+        return JsonResponse({"is_admin": False})
```
juliasaveliff: moved this johnson from models to users where it belongs

benauriemma: Did you just.......... `if true return true`

benauriemma: Also I imagine it defaults to `status=HTTPStatus.OK` but idk if we want to say that explicitly?

benauriemma: Did you just.......... `if true return true`

benauriemma: Also I imagine it defaults to `status=HTTPStatus.OK` but idk if we want to say that explicitly?

juliasaveliff: reverse uno card

PR #205, opened 2020-02-18
haha making some small changes

brendancheng20:
- [x] datacenter table entries no longer clickable since there isn't a detailView

- [x] Front end shows power options button if asset is powered, no functionality yet tho

### DIFF
--- a/src/components/elementView/elementTable.tsx
+++ b/src/components/elementView/elementTable.tsx
@@ -98,7 +98,7 @@ interface ElementTableProps {
 class ElementTable extends React.Component<
   ElementTableProps & RouteComponentProps,
   ElementTableState
-> {
+  > {
   public state: ElementTableState = {
     page_type: 10,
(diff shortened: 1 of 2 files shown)


src/components/elementView/elementTable.tsx
```diff
@@ -800,9 +804,12 @@ class ElementTable extends React.Component<
               {this.state.items && this.state.items.length > 0 ? (
                 <tbody>
                   {this.state.items.map((item: ElementObjectType) => {
+                    if (isAssetObject(item)) {
+                      console.log(item)
+                    }
                     return (
                       <tr
-                        onClick={() => {
+                        onClick={this.props.type === "datacenters" ? () => { } : () => {
```
clairecfu: Can you use the enum just for consistency? `ElementTypes.DATACENTER`. Also while you're at it, can you disable clicking for users as well?

PR #206, opened 2020-02-18
Defining new types for assets

clairecfu:
- [x] AssetObject -> AssetObjectOld
- [x] AssetInfoObject -> ShallowAssetObjectOld
- [x] the UPDATED AssetInfoObject -> ShallowAssetObject

The new types have not been implemented anywhere -> i am using them on a diff branch for asset forms, but we should start using the new types for new features being developed/migrate away from the old objects.

### DIFF
--- a/src/utils/utils.tsx
+++ b/src/utils/utils.tsx
@@ -10,13 +10,44 @@ export enum ElementType {
   USER = "users",
   DATACENTER = "datacenters"
 }
-export interface AssetObject extends ElementObject {
+export enum PowerSide {
+  LEFT = "L",
+  RIGHT = "R"
+}
+export interface AssetObjectOld extends ElementObject {
   hostname: string;
   rack_position: string;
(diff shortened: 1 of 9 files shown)


src/utils/utils.tsx
```diff
@@ -10,14 +10,45 @@ export enum ElementType {
...
+}
+export interface AssetObject extends ParentAssetObject {
+  model: ModelObject;
+  rack: RackObject;
+}
+
+interface ParentAssetObject extends ElementObject {
+  hostname: string;
+  rack_position: string;
+  mac_addresses: { [port: string]: string };
```
brendancheng20: What is this syntax? Mostly just asking out of curiosity, I've never seen it before

clairecfu: its basically a dict syntax, so mac addresses has keys of type string, and values of type string.

PR #207, opened 2020-02-18
adding power connections on asset add and modify and returning them on get detail

benauriemma:


rackcity/views/asset_views.py
```diff
@@ -194,17 +187,24 @@ def asset_add(request):  # need to make network and power connections here
             except MacAddressException as error:
                 failure_message += "Some mac addresses couldn't be saved. " + \
                     str(error)
-                return JsonResponse(
-                    {"failure_message": failure_message},
-                    status=HTTPStatus.BAD_REQUEST,
+            try:
+                save_power_connections(
```
benauriemma: If mac addresses fail, keep track of the errors but still try to save power connections

rackcity/views/asset_views.py
```diff
@@ -234,6 +237,54 @@ def save_mac_addresses(asset_data, asset_id):
         raise MacAddressException(failure_message)
 
 
+def save_power_connections(asset_data, asset_id):
```
benauriemma: I feel like this function and the mac address equivalent could live somewhere better, but I'm also not sure we have a better place currently. Might be part of a larger discussion

rackcity/views/rackcity_utils.py
```diff
@@ -177,6 +177,11 @@ def __init__(self, *args, **kwargs):
         Exception.__init__(self, *args, **kwargs)
 
 
+class PowerConnectionException(Exception):
```
benauriemma: I want to put custom exceptions in their own file at some point but it's low priority

rackcity/api/serializers/asset_serializers.py
```diff
@@ -50,8 +53,35 @@ class Meta:
...
+        except ObjectDoesNotExist:
+            return
+        power_connections = {}
+        for port in ports:
+            if port.power_connection:
+                power_connections[port.port_name] = {
+                    "left_right": port.power_connection.left_right,
+                    "port_number": port.power_connection.port_number
+                }
+        return power_connections  # YOU ARE HERE! Test with one that has connections
```
benauriemma: I WAS HERE! And I did test with one that has connections. I forgot to delete the comment but I deleted it in my next PR

rackcity/views/rackcity_utils.py
juliasaveliff: good idea

rackcity/views/asset_views.py
```diff
@@ -321,12 +373,24 @@ def asset_modify(request):
...
             failure_message += "Some mac addresses couldn't be saved. " + \
                 str(error)
+        try:
+            save_power_connections(
+                asset_data=data,
+                asset_id=existing_asset.id
+            )
+        except PowerConnectionException as error:
+            failure_message += "Some power connections couldn't be saved. " + \
+                str(error)
```
juliasaveliff: I don't think we want to throw unknown `str(error)` into the failure message because this is stuff that might look weird to the user. I'm fine with merging this as it is, but that will just mean more work for when we try to fix up all the errors. Up to you!

rackcity/views/asset_views.py
```diff
@@ -321,12 +373,24 @@ def asset_modify(request):
...
+        if failure_message:
             return JsonResponse(
                 {"failure_message": failure_message},
                 status=HTTPStatus.BAD_REQUEST,
             )
         else:
-            return HttpResponse(status=HTTPStatus.OK)
+            return JsonResponse(
```
juliasaveliff: might get merge conflicts here with the logs I added

rackcity/views/asset_views.py
juliasaveliff: Nvm I see these errors are failure messages you construct in save_power_connections. Worth usability testing later on what these will look like in action but looks good for now

rackcity/api/serializers/asset_serializers.py
```diff
@@ -50,8 +53,35 @@ class Meta:
...
+    def get_mac_addresses(self, asset):
+        try:
+            ports = NetworkPort.objects.filter(asset=asset.id)
+        except ObjectDoesNotExist:
+            return
+        mac_addresses = {}
+        for port in ports:
+            if port.mac_address:
+                mac_addresses[port.port_name] = port.mac_address
+        return mac_addresses
```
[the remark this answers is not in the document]
benauriemma: Just for you

rackcity/views/asset_views.py
benauriemma: I tried all the error messages out in postman and the ones I wrote myself look pretty and informative to me.  Have been trying to keep those clean as much as possible

PR #208, opened 2020-02-18
network connections api layer

benauriemma:


benauriemma:
This is finished, but I pulled from my other PR so may be easier to hold off on reviewing until that's merged

rackcity/models/network_port.py
```diff
@@ -25,6 +26,46 @@ class NetworkPort(models.Model):
...
+    def create_network_connection(self, destination_port):
+        print("creating connection from "+str(self.id) +
+              " to  "+str(destination_port.id))
+        if (
+            destination_port.network_connection
+            and destination_port.network_connection != self
+        ):
+            from rackcity.views.rackcity_utils import (
+                NetworkConnectionException
+            )
```
benauriemma: Avoiding circular import stuff again

rackcity/views/asset_views.py
```diff
@@ -164,49 +170,66 @@ def asset_add(request):  # need to make network and power connections here
     failure_message = ""
     if 'id' in data:
         failure_message += "Don't include id when adding an asset. "
-
+        return JsonResponse(
+            {"failure_message": failure_message},
+            status=HTTPStatus.BAD_REQUEST,
+        )
```
benauriemma: Changing this to fail early

rackcity/views/asset_views.py
```diff
@@ -164,49 +170,66 @@ def asset_add(request):  # need to make network and power connections here
...
     serializer = AssetSerializer(data=data)
     if not serializer.is_valid(raise_exception=False):
         failure_message += str(serializer.errors)
-    if failure_message == "":
```
benauriemma: A lot of this big chunk is just because I un-indented now that we're failing early.  Sorry it makes it hard to read

rackcity/views/asset_views.py
```diff
@@ -164,49 +170,66 @@ def asset_add(request):  # need to make network and power connections here
...
+            asset_id=asset.id
+        )
+    except NetworkConnectionException as error:
+        warning_message += "Some network connections couldn't be saved. " + \
+            str(error)
+    if warning_message:
+        return JsonResponse(
+            {"warning_message": warning_message},
+            status=HTTPStatus.OK,
+        )
```
benauriemma: 🚨🚨🚨 flagging to @clairecfu !!! I'm open to discussing other ways to do this but essentially this is what we talked about that there's a new case we have to handle in forms where the asset was created or modified successfully (form should not stay open) but we failed to create all mac addresses, power connections, and/or network connections. The backend returns a very descriptive message in these cases but we can't just throw it away because it was a status=200. I looked into whether there's a warning like HTTP status (personally I think it would be called `status=HTTPStatus.OK_BUT`) but it doesn't exist.  We could agree on a unique status for these cases though

rackcity/views/asset_views.py
```diff
@@ -245,6 +268,82 @@ def save_mac_addresses(asset_data, asset_id):
         raise MacAddressException(failure_message)
 
 
+def save_network_connections(asset_data, asset_id):
```
benauriemma: What will happen first? 1000 commits or 1000 lines in asset_views? Flagged on the last PR, I think this and the mac/power equivalents should live elsewhere, but I'm not convinced a better place currently exists

rackcity/views/rackcity_utils.py
```diff
@@ -311,3 +318,13 @@ def get_filter_arguments(data):
...
+
+
+def close_old_connections_decorator(func):
+    @functools.wraps(func)
+    def wrapper_decorator(*args, **kwargs):
+        close_old_connections()
+        value = func(*args, **kwargs)
+        close_old_connections()
+        return value
+    return wrapper_decorator
```
benauriemma: This is the start of an experimental fix for the database connections issue - it fixed itself before I could test, but was going to leave this here in case it comes up again and someone wants to try commenting out the decorators  I added to asset views

benauriemma: If this is the route we would want to go (I don't have high conviction that it is), I think it would ultimately make sense to roll this decorator in the @api_view decorator so we don't have to repeat it every time

rackcity/models/network_port.py
```diff
@@ -25,6 +26,46 @@ class NetworkPort(models.Model):
...
+                ":" +
+                destination_port.port_name +
+                "' is already connected to port '" +
+                destination_port.network_connection.asset.hostname +
+                ":" +
+                destination_port.network_connection.port_name +
+                "'. "
+            )
+        if self.network_connection:
+            self.network_connection.network_connection = None
```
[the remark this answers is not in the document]
juliasaveliff: if this is on network connection create, why do we need to set this to None (wouldn't it default to Null)? I understand why we would need to do this if we're deleting a connection so that the other uni-directional onetoone connection gets deleted too, but why on create?

rackcity/api/serializers/asset_serializers.py
```diff
@@ -80,7 +82,24 @@ def get_power_connections(self, asset):
...
+            ports = NetworkPort.objects.filter(asset=asset.id)
+        except ObjectDoesNotExist:
+            return
+        network_connections = []
+        for port in ports:
+            if port.network_connection:
+                destination_port = port.network_connection
+                network_connection_serialized = {
+                    "source_port": port.port_name,
+                    "destination_hostname": destination_port.asset.hostname,
```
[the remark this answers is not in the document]
benauriemma: Yep - bulk committee realized today that it's not in writing so we'll likely request it in writing but he definitely said it in class

rackcity/models/network_port.py
benauriemma: The link pointing from source port to destination port is about to be changed, so we need to remove the link pointing from destination port to source port so we're not left with a dangling half connection. Do you think this should be commented in the code?

juliasaveliff: per our slack discussion, perhaps replace this code with a call to `delete_network_connection` is it's clear what is happening

rackcity/models/network_port.py
```diff
@@ -25,6 +26,46 @@ class NetworkPort(models.Model):
         blank=True,
     )
 
+    def create_network_connection(self, destination_port):
```
juliasaveliff: per our slack conversation, perhaps rename `network_connection` field to `connected_port` so it's obvious that this entity is another NetworkPort, and not some sort of connection object

benauriemma: I like this - changed to reflect

rackcity/models/network_port.py
benauriemma: Done. This helps a lot

PR #211, opened 2020-02-19
all backend changes for ev2 bulk import and export

benauriemma:


rackcity/views/asset_views.py
```diff
@@ -547,29 +549,42 @@ def asset_delete(request):  # need to delete network and power connections here
...
     hostnames_in_import = set()
-    for asset_data in asset_datas:
-        if (
-            'vendor' not in asset_data
-            or 'model_number' not in asset_data
-        ):
-            return JsonResponse(
-                {"failure_message": "Asset records must include 'vendor' and 'model_number'. "},
-                status=HTTPStatus.BAD_REQUEST
-            )
```
benauriemma: We already validated columns

rackcity/views/asset_views.py
```diff
@@ -586,15 +601,25 @@ def asset_bulk_upload(request):  # need to make network and power connections he
         asset_data['model'] = model.id
         del asset_data['vendor']
         del asset_data['model_number']
-        if 'rack' not in asset_data:
```
benauriemma: We already validated columns

rackcity/api/serializers/asset_serializers.py
```diff
@@ -121,20 +127,70 @@ class BulkAssetSerializer(serializers.ModelSerializer):
...
+            'power_port_connection_2'
         )
 
     def get_rack(self, asset):
         return asset.rack.row_letter + str(asset.rack.rack_num)
+
+    def get_power_port_connection_1(self, asset):
+        return self.power_port_connection(asset, port_number=1)
+
+    def get_power_port_connection_2(self, asset):
```
brendancheng20: Do we want to generalize this somehow? What if there are more than 2 power ports

benauriemma: The bulk format spec only has two columns for this, any additional connections would have to be specified manually

benauriemma:
Important!! A looot of changes in asset_views but it'll be collapsed by default because it's a big boy

rackcity/api/serializers/asset_serializers.py
```diff
@@ -110,24 +116,74 @@ class BulkAssetSerializer(serializers.ModelSerializer):
...
+    if bulk_asset_data['power_port_connection_2']:
+        power_connections["2"] = {
+            "left_right": bulk_asset_data['power_port_connection_2'][:1],
+            "port_number": int(bulk_asset_data['power_port_connection_2'][1:])
+        }
+    bulk_asset_data['power_connections'] = power_connections
+    del bulk_asset_data['power_port_connection_1']
+    del bulk_asset_data['power_port_connection_2']
+    if not bulk_asset_data['asset_number']:
+        del bulk_asset_data['asset_number']
```
brendancheng20: What's the purpose of using this del keyword? Are we just deleting that field off of bulk_asset_data? How does that help us?

brendancheng20: Apologies for incompetence, am only like 3/4 stack over here I guess

rackcity/api/serializers/asset_serializers.py
```diff
@@ -110,24 +116,74 @@ class BulkAssetSerializer(serializers.ModelSerializer):
...
+            "port_number": int(bulk_asset_data['power_port_connection_1'][1:])
+        }
+    if bulk_asset_data['power_port_connection_2']:
+        power_connections["2"] = {
+            "left_right": bulk_asset_data['power_port_connection_2'][:1],
+            "port_number": int(bulk_asset_data['power_port_connection_2'][1:])
+        }
+    bulk_asset_data['power_connections'] = power_connections
+    del bulk_asset_data['power_port_connection_1']
+    del bulk_asset_data['power_port_connection_2']
```
juliasaveliff: need to ensure we never get a KeyError on del. are we just assuming the bulk import file will be formatted correctly, will all specified fields included but some might be null?

rackcity/api/serializers/asset_serializers.py
juliasaveliff: Ben showed me this the other day, it's part of his "massage". The bulk import will send this data in the format style
```
{
'power_port_connection_1': L19,
'power_port_connection_2': R19
}
```
but we want to look like this
```
{
"1": {"left_right": L, "port_number": 19},
"2": {"left_right": R, "port_number": 19}
}
```
so we can directly serialize the incoming data into one of our backend objects. So he's just reformatting what we receive here

rackcity/api/serializers/it_model_serializers.py
```diff
@@ -71,10 +75,59 @@ class Meta:
...
+            network_ports.append(str(port_number))
+    bulk_model_data['network_ports'] = network_ports
+    del bulk_model_data['network_port_name_1']
+    del bulk_model_data['network_port_name_2']
+    del bulk_model_data['network_port_name_3']
+    del bulk_model_data['network_port_name_4']
+    bulk_model_data['num_power_ports'] = bulk_model_data['power_ports']
+    del bulk_model_data['power_ports']
+    bulk_model_data['memory_gb'] = bulk_model_data['memory']
+    del bulk_model_data['memory']
```
juliasaveliff: same question about KeyError on del

rackcity/api/serializers/asset_serializers.py
juliasaveliff: JK I see that you validate all column names in assets views, so that answers my question

rackcity/api/serializers/it_model_serializers.py
juliasaveliff: answered above

PR #213, opened 2020-02-19
Put username in navigation bar

juliasaveliff:


djreact/settings.py
```diff
@@ -78,7 +78,7 @@
 #     }
 # }
 DATABASES = {}
-DATABASES['default'] = dj_database_url.config(conn_max_age=600)
+DATABASES['default'] = dj_database_url.config()
```
juliasaveliff: hoping this helps to fix the database connections error

PR #215, opened 2020-02-19
Connect grant/revoke admin endpoints to buttons

juliasaveliff:
- When click "Remove Admin" or "Add Admin" on a user table, confirmation dialog appears, then toast displays outcome, and table updates
- Also fixed a bug coming from LOGGING delete actions that was messing with delete actions

### DIFF
--- a/rackcity/utils/log_utils.py
+++ b/rackcity/utils/log_utils.py
@@ -81,7 +81,7 @@ def log_delete(user, element_type, element_name):
         element_name + ":",
         user.username,
         Action.DELETE.value,
-        element_type,
+        element_type.value,
         element_name
     ])
     log = Log(
--- a/src/components/elementView/elementTable.tsx
+++ b/src/components/elementView/elementTable.tsx
@@ -51,8 +51,10 @@ import {
   deleteDatacenter,
   modifyModel,
   modifyAsset,
-  modifyDatacenter
+  modifyDatacenter,
+  ElementTableOpenAlert
 } from "./elementUtils";
+import { ELEVATION_0 } from "@blueprintjs/core/lib/esm/common/classes";
 
 interface ElementTableState {
   items: Array<ElementObjectType>;
@@ -66,7 +68,8 @@ interface ElementTableState {
   fields: Array<string>;
   isEditFormOpen: boolean;
   editFormValues: ElementObjectType;
-  isDeleteOpen: boolean;
+  openAlert: ElementTableOpenAlert;
+  selected_userid?: string;
 }
 // var console: any = {};
 // console.log = function() {};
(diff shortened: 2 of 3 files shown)


rackcity/utils/log_utils.py
```diff
@@ -81,7 +81,7 @@ def log_delete(user, element_type, element_name):
         element_name + ":",
         user.username,
         Action.DELETE.value,
-        element_type,
+        element_type.value,
```
juliasaveliff: Bug fix from log (need the enum value which is a str)

src/components/elementView/elementTable.tsx
```diff
@@ -64,7 +67,8 @@ interface ElementTableState {
   fields: Array<string>;
   isEditFormOpen: boolean;
   editFormValues: ElementObjectType;
-  isDeleteOpen: boolean;
+  openAlert: ElementTableOpenAlert;
```
juliasaveliff: other option would have been to have 3 bools on state (isDeleteOpen, isGrantAdminOpen, isRevokeAdminOpen) to handle all alerts, which felt gross!

src/components/elementView/elementTable.tsx
```diff
@@ -691,12 +760,32 @@ class ElementTable extends React.Component<
...
         >
           <p>Are you sure you want to delete?</p>
         </Alert>
+        <Alert
+          cancelButtonText="Cancel"
+          confirmButtonText="Confirm"
+          intent="danger"
+          isOpen={this.state.openAlert === ElementTableOpenAlert.GRANT_ADMIN}
+          onCancel={this.handleGrantAdminCancel}
+          onConfirm={this.handleGrantAdmin}
```
juliasaveliff: this nugget right here is why I created the `selected_userid` field, the method `handleGrantAdmin` needs to know which user was selected when this button was clicked. if there is a better way to do this that i'm missing that doesn't required the selected user/row field on state, I can change this!

src/components/elementView/elementTable.tsx
```diff
@@ -863,44 +952,44 @@ class ElementTable extends React.Component<
...
-                                onClick={(event: any) => {
-                                  this.handleDeleteButtonClick(item);
-                                  event.stopPropagation();
-                                }}
-                              />
-                              {this.props.isAdmin &&
-                              isAssetObject(item) &&
-                              item.rack.is_network_controlled ? (
+                            this.props.type !== ElementType.USER ? (
+                              <div className="inline-buttons">
```
juliasaveliff: I don't know why git thinks I changed all the code around here (maybe automatic formatting changes???) but I don't think any of the code beyond this point actually changed

src/components/elementView/elementUtils.tsx
```diff
@@ -12,6 +12,13 @@ export interface ITableSort {
   id: string;
 }
 
+export enum ElementTableOpenAlert {
+  NONE = "none",
+  DELETE = "delete",
+  GRANT_ADMIN = "grant_admin",
+  REVOKE_ADMIN = "revoke_admin"
+}
+
```
clairecfu: i LOVE THIS!!!!!

src/components/elementView/elementTable.tsx
clairecfu: trueeee

PR #216, opened 2020-02-19
No one rack should have all that power

brendancheng20:
set up the endpoint for getting power information for a given asset

### DIFF
--- a/rackcity/views/pdu_views.py
+++ b/rackcity/views/pdu_views.py
@@ -1,35 +1,17 @@
-# from django.http import HttpResponse, JsonResponse
 from django.http import JsonResponse
 from rackcity.models import (
-    Rack,
     Asset
 )
 from rackcity.api.serializers import (
     serialize_power_connections,
 )
-# from rackcity.utils.log_utils import (
-#     log_action,
-#     log_bulk_import,
-#     log_delete,
-#     Action,
-#     ElementType,
-# )
 from rest_framework.decorators import permission_classes, api_view
-from rest_framework.permissions import IsAuthenticated, IsAdminUser
-# from rest_framework.parsers import JSONParser
+from rest_framework.permissions import IsAuthenticated
 from http import HTTPStatus
-from rackcity.views.rackcity_utils import (
-    validate_asset_location,
-    validate_location_modification,
-    no_infile_location_conflicts,
-    records_are_identical,
-    get_sort_arguments,
-    get_filter_arguments,
-)
 import re
 import requests
 
-pdu_url = 'http://hyposoft-mgt.colab.duke.edu:8000/pdu.php?pdu=hpdu-rtp1-' # Need to specify rack + side in request, e.g. for A1 left, use A01L
+pdu_url = 'http://hyposoft-mgt.colab.duke.edu:8005/pdu.php?pdu=hpdu-rtp1-'  # Need to specify rack + side in request, e.g. for A1 left, use A01L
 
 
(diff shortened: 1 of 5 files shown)


rackcity/views/pdu_views.py
```diff
@@ -0,0 +1,74 @@
...
+from rackcity.api.serializers import (
+    serialize_power_connections,
+)
+# from rackcity.utils.log_utils import (
+#     log_action,
+#     log_bulk_import,
+#     log_delete,
+#     Action,
+#     ElementType,
+# )
```
benauriemma: Can you delete commented imports

rackcity/views/pdu_views.py
```diff
@@ -0,0 +1,74 @@
...
+# from rest_framework.parsers import JSONParser
+from http import HTTPStatus
+from rackcity.views.rackcity_utils import (
+    validate_asset_location,
+    validate_location_modification,
+    no_infile_location_conflicts,
+    records_are_identical,
+    get_sort_arguments,
+    get_filter_arguments,
+)
```
benauriemma: And any unused ones as well

rackcity/views/pdu_views.py
```diff
@@ -0,0 +1,74 @@
...
+import re
+import requests
+
+pdu_url = 'http://hyposoft-mgt.colab.duke.edu:8000/pdu.php?pdu=hpdu-rtp1-' # Need to specify rack + side in request, e.g. for A1 left, use A01L
```
benauriemma: Ours is 8005 - do we need to use 8000 to test?

brendancheng20: Using 8000 for testing purposes bc 8005 doesn't work when I run locally. I suppose we should change it now and see what happens on dev

rackcity/views/pdu_views.py
brendancheng20: Will do, totally forgot about this stuff

rackcity/views/pdu_views.py
```diff
@@ -0,0 +1,74 @@
...
+            {"failure_message": failure_message},
+            status=HTTPStatus.BAD_REQUEST
+        )
+    port_info = serialize_power_connections(asset)
+
+    # get string parameter representing rack number (i.e. A01<L/R>)
+    rack_str = str(asset.rack.row_letter)
+    if (asset.rack.rack_num / 10 < 1):
+        rack_str = rack_str + "0"
+    rack_str = rack_str + str(asset.rack.rack_num)
```
benauriemma: I think you can do like a %s thing to format the number into the string but this is fine

rackcity/views/pdu_views.py
```diff
@@ -0,0 +1,74 @@
...
+
+    power_status = dict()
+    for port in port_info:
```
benauriemma: Could you rename `port_info` to `power_connections` and this to `power_connection in power_connections`? I was a little confused by the name

rackcity/views/pdu_views.py
```diff
@@ -0,0 +1,74 @@
...
+            "power_connections": port_info,
+            "power_status": power_status
+        },
+        status=HTTPStatus.OK
+    )
+
+
+def regex_power_status(html, port):
+    status_pattern = re.search('<td>'+str(port)+'<td><span.+(ON|OFF)', html)
+    return status_pattern.groups(1)
```
benauriemma: The meat of this is an absolute beauty

PR #218, opened 2020-02-20
FIXED model view

clairecfu:


src/components/elementView/detailedView/modelView/modelView.tsx
```diff
@@ -127,7 +127,7 @@ export class ModelView extends React.PureComponent<
     console.log(this.state.assets);
     let params: any;
     params = this.props.match.params;
-    if (this.state.model === undefined) {
+    if (Object.keys(this.state.model).length === 0) {
```
clairecfu: bug fix!! refactored the code earlier so that model is never actually undefined

src/components/elementView/detailedView/rackView/rackView.tsx
```diff
@@ -166,13 +163,6 @@ class RackView extends React.PureComponent<
   };
 
   public render() {
-    // if (this.props.location.state.length === 0) {
-    //   this.getRackRange(this.props.token);
-    // }
-    // const racks: Array<RackResponseObject> =
-    //   this.props.location && this.props.location.state
-    //     ? (this.props.location.state as Array<RackResponseObject>)
-    //     : [];
```
clairecfu: removed comments

PR #219, opened 2020-02-20
Welcome, username message bug fixes

juliasaveliff:
- Now "Welcome, username" is not displayed until the username has loaded (was previously flashing "Welcome, undefined" before switching to something like "Welcome, jms247")
- Username is cleared when user logs out, so welcome message properly updates if the user logs back in to a different account

### DIFF
--- a/src/components/navigation/navigation.tsx
+++ b/src/components/navigation/navigation.tsx
@@ -53,7 +53,7 @@ export class Navigation extends React.Component<NavigationPropsAll, NavigationSt
   }
 
   public render() {
-    if (!this.state.username) {
+    if (this.props.isAuthenticated && !this.state.username) {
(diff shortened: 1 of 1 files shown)


src/components/navigation/navigation.tsx
```diff
@@ -47,9 +47,13 @@ export class Navigation extends React.Component<NavigationPropsAll, NavigationSt
...
       });
   }
+  clearUsernameAndLogout() {
+    this.setState({ username: undefined })
+    this.props.logout()
+  }
 
   public render() {
-    if (!this.state.username) {
+    if (this.props.isAuthenticated && !this.state.username) {
```
juliasaveliff: was getting some 401 on /who-am-i when nav bar is rendered for unlogged in user

PR #220, opened 2020-02-20
Fail if try to delete netid user

juliasaveliff:
In the ev2 spec: "Administrators can create and delete “local” (non-NetID) user accounts."

### DIFF
--- a/rackcity/utils/user_utils.py
+++ b/rackcity/utils/user_utils.py
@@ -1,0 +1,8 @@
+def is_netid_user(user):
+    """
+    Returns true if user is Duke SSO authenticated NetID user.
+    Determined by whether the user has a non-empty password, since
+    NetID users are authenticated by access token and their associated
+    passwords are empty strings.
+    """
+    return (not user.password)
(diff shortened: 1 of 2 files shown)


rackcity/utils/user_utils.py
```diff
@@ -0,0 +1,8 @@
+def is_netid_user(user):
+    """
+    Returns true if user is Duke SSO authenticated NetID user.
+    Determined by whether the user has a non-empty password, since
+    NetID users are authenticated by access token and their associated
+    passwords are empty strings.
+    """
+    return (not user.password)
```
juliasaveliff: learned some fun facts about passwords (hashed + salted) on django user model. they are saved as `''` not `null` if a user is created without a password (which we do for netid users). additional fun facts about python and strings: `not 'non-empty string' = False` and `not '' = True`, while `'non-empty string' is None = False` and `'' is None = True`

juliasaveliff: I unfortunately accidentally deleted my roommates from our system while testing the differences. plz don't tell them

clairecfu: this feels like a riddle. am too brain dead post midterm to understand

PR #223, opened 2020-02-20
Set up power backend endpoints

brendancheng20:
- Power cycle

- Power on/off

- Validation of power states (i.e. if power is already on when power-on request is sent, alert user and do nothing)

### DIFF
--- a/rackcity/api/urls.py
+++ b/rackcity/api/urls.py
@@ -49,5 +49,7 @@ urlpatterns = [
     path('users/pages', views.user_page_count),
     path('users/grant-admin', views.user_grant_admin),
     path('users/revoke-admin', views.user_revoke_admin),
-    path('power/get-state/<int:id>', views.power_status)
+    path('power/get-state/<int:id>', views.power_status),
+    path('power/mask-on', views.power_on),
+    path('power/mask-off', views.power_off)
 ]
--- a/rackcity/api/urls.py
+++ b/rackcity/api/urls.py
@@ -51,5 +51,6 @@ urlpatterns = [
     path('users/revoke-admin', views.user_revoke_admin),
     path('power/get-state/<int:id>', views.power_status),
     path('power/mask-on', views.power_on),
-    path('power/mask-off', views.power_off)
+    path('power/mask-off', views.power_off),
+    path('power/cycle', views.power_cycle)
 ]
--- a/rackcity/views/pdu_views.py
+++ b/rackcity/views/pdu_views.py
@@ -1,17 +1,20 @@
-from django.http import JsonResponse
+from django.http import HttpResponse, JsonResponse
 from rackcity.models import (
     Asset
 )
 from rackcity.api.serializers import (
     serialize_power_connections,
 )
+from rest_framework.parsers import JSONParser
 from rest_framework.decorators import permission_classes, api_view
-from rest_framework.permissions import IsAuthenticated
+from rest_framework.permissions import IsAuthenticated, IsAdminUser
 from http import HTTPStatus
 import re
 import requests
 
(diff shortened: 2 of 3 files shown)


rackcity/api/urls.py
```diff
@@ -49,5 +49,8 @@
     path('users/pages', views.user_page_count),
     path('users/grant-admin', views.user_grant_admin),
     path('users/revoke-admin', views.user_revoke_admin),
-    path('power/get-state/<int:id>', views.power_status)
+    path('power/get-state/<int:id>', views.power_status),
+    path('power/mask-on', views.power_on),
+    path('power/mask-off', views.power_off),
```
juliasaveliff: I am loling

rackcity/views/pdu_views.py
```diff
@@ -51,6 +55,110 @@ def power_status(request, id):
...
+        )
+    power_connections = serialize_power_connections(asset)
+    # Check power is off
+    html = requests.get(pdu_url + get_pdu + get_pdu_status_ext(asset, str(power_connections[data['power_port_number']]['left_right'])))
+    power_status = regex_power_status(html.text, power_connections[data['power_port_number']]['port_number'])[0]
+    if power_status == "ON":
+        return JsonResponse(
+            {"data": "Port is already on"},
+            status=HTTPStatus.OK
+        )
```
juliasaveliff: I saw in another PR that @benauriemma started returning responses like this as 
```
     return JsonResponse(
            {"warning_message": warning_message},
            status=HTTPStatus.OK,
        )
```
not sure which path we want to stick with but I would either rename `data` to `success_message` or `warning_message` to be consistent with the other response messages ?

rackcity/views/pdu_views.py
```diff
@@ -51,6 +55,110 @@ def power_status(request, id):
...
+        asset = Asset.objects.get(id=data['id'])
+    except Asset.DoesNotExist:
+        failure_message = "No model exists with id=" + str(id)
+        return JsonResponse(
+            {"failure_message": failure_message},
+            status=HTTPStatus.BAD_REQUEST,
+        )
+    power_connections = serialize_power_connections(asset)
+    # Check power is off
+    html = requests.get(pdu_url + get_pdu + get_pdu_status_ext(asset, str(power_connections[data['power_port_number']]['left_right'])))
```
juliasaveliff: it might be worth checking that the keys `'id'` and `'power_port_number'` are in data and failing early if not, and maybe also make sure the power_port_number is a valid port on the asset's model. this is mostly helpful for controlling the error messages that get shown to the user so no key errors get tangled up and throw a 500

rackcity/views/pdu_views.py
```diff
@@ -51,6 +55,110 @@ def power_status(request, id):
...
+    power_status = regex_power_status(html.text, power_connections[data['power_port_number']]['port_number'])[0]
+    if power_status == "ON":
+        return JsonResponse(
+            {"data": "Port is already on"},
+            status=HTTPStatus.OK
+        )
+    toggle_power(asset, data['power_port_number'], "on")
+    return HttpResponse(status=HTTPStatus.OK)
```
juliasaveliff: do we want to send a success_message to the front end to display? or will the view in the front end automatically update?

rackcity/views/pdu_views.py
```diff
@@ -51,6 +55,110 @@ def power_status(request, id):
...
+    except Asset.DoesNotExist:
+        failure_message = "No model exists with id=" + str(id)
+        return JsonResponse(
+            {"failure_message": failure_message},
+            status=HTTPStatus.BAD_REQUEST,
+        )
+    power_connections = serialize_power_connections(asset)
+    # Check power is off
+    html = requests.get(pdu_url + get_pdu + get_pdu_status_ext(asset, str(power_connections[data['power_port_number']]['left_right'])))
+    power_status = regex_power_status(html.text, power_connections[data['power_port_number']]['port_number'])[0]
```
juliasaveliff: regex 👑

PR #224, opened 2020-02-21
Asset Forms, pt 2.

clairecfu:
- [x] specifying network connections for assets
- [x] switch between datacenters  (TODO: ERROR CHECKING FOR THIS ON BACKEND)
- [x] does not allow modification of an asset's model

to be done in another PR:

- [ ] convert empty strings to undefined for asset name and hostname
- [ ] mac address validation 
- [ ] connect power port drop down info to backend endpoints

### DIFF
--- a/src/forms/assetForm.tsx
+++ b/src/forms/assetForm.tsx
@@ -801,6 +801,8 @@ class AssetForm extends React.Component<AssetFormProps, AssetFormState> {
               >
                 <Button
                   rightIcon="caret-down"
+                  disabled={!isNullOrUndefined(this.initialState.model)}
+
                   text={
                     this.state.values.model
                       ? this.state.values.model.vendor +
--- a/src/forms/assetForm.tsx
+++ b/src/forms/assetForm.tsx
@@ -45,7 +45,8 @@ import {
   renderModelItem,
   renderRackItem,
   renderStringItem,
-  StringSelect
+  StringSelect,
+  AssetSelect
 } from "./formUtils";
 
 //TO DO : add validation of types!!!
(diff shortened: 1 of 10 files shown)


src/forms/assetForm.tsx
```diff
@@ -97,20 +111,25 @@ class AssetForm extends React.Component<AssetFormProps, AssetFormState> {
...
     users: [],
-    currModel: {} as ModelObject,
+    assets: [],
     //TODO, call endpoint, don't hard code
     power_ports: {
       left_suggest: "12",
       left_available: ["1", "2", "12"],
       right_suggest: "12",
       right_available: ["1", "2", "12", "13"]
     },
```
clairecfu: only non function part - available/suggested power ports are still hard coded

src/utils/utils.tsx
```diff
@@ -31,13 +31,14 @@ export interface AssetObject extends ParentAssetObject {
...
+  asset_number: string;
   hostname: string;
   rack_position: string;
   mac_addresses: { [port: string]: string };
   network_connections: Array<NetworkConnection>;
   power_connections: { [port: string]: PowerConnection };
-  owner?: string;
-  comment?: string;
+  owner: string;
+  comment: string;
```
clairecfu: removing these as ` |undefined` fields in the frontend -> this is not and should not be direct mapping between a required/optional field in the backend

src/forms/assetForm.tsx
```diff
@@ -259,32 +291,76 @@ class AssetForm extends React.Component<AssetFormProps, AssetFormState> {
...
+    ).then(res => {
+      let assetsWithHostname: Array<AssetObject> = res as Array<AssetObject>;
+      assetsWithHostname = assetsWithHostname.filter(asset => {
+        if (
+          asset.hostname === "" ||
+          asset.hostname === this.state.values.hostname
+        ) {
+          return false;
+        }
+        return true;
```
juliasaveliff: nice catch

PR #226, opened 2020-02-21
Add sort, filter to users table

juliasaveliff:


rackcity/api/serializers/user_serializers.py
```diff
@@ -14,6 +14,8 @@ def custom_signup(self, request, user):
 
 
 class UserSerializer(serializers.ModelSerializer):
+    is_admin = serializers.BooleanField(source='is_staff')
+
```
juliasaveliff: Renamed this from `is_staff` to `is_admin` in backend & front end

rackcity/views/user_views.py
```diff
@@ -68,7 +72,7 @@ def netid_login(request):
 
 @api_view(['POST'])
 @permission_classes([IsAdminUser])
-def user_list(request):
+def user_many(request):
```
juliasaveliff: renamed to be consistent

rackcity/views/user_views.py
```diff
@@ -99,7 +103,28 @@ def user_list(request):
...
+            {"failure_message": "Filter error: " + str(error)},
+            status=HTTPStatus.BAD_REQUEST
+        )
+    for filter_arg in filter_args:
+        users_query = users_query.filter(**filter_arg)
+
+    try:
+        sort_args = get_sort_arguments(request.data)
+        if len(sort_args) == 0:
+            sort_args = ['username']
```
juliasaveliff: add default sort if none is specified to fix the weird behavior where modifying the admin permission of a user yeets them to the last row

src/components/userView/user.scss
```diff
@@ -0,0 +1,9 @@
+.user-view {
+    padding-left: 20px;
+    padding-right: 20px;
+    margin-bottom: 20px;
+    width: 100%;
+    margin: 10pt;
+    overflow: clip;
+    width: calc(100% - 200pt) !important;
+  }
```
juliasaveliff: beautifying users page

src/components/elementView/elementTable.tsx
```diff
@@ -512,7 +512,8 @@ class ElementTable extends React.Component<
         col !== "comment" &&
         col !== "power_connections" &&
         col !== "mac_addresses" &&
-        col !== "network_connections"
+        col !== "network_connections" &&
+        col !== "is_admin"
```
juliasaveliff: made it so we don't show is_admin column in the table. this is a bool so its values never appear in the column. I think it's redundant anyway since the grant/revoke admin button right next to it makes it obvious what the permission is currently

rackcity/api/serializers/user_serializers.py
clairecfu: yay

src/components/elementView/elementTable.tsx
clairecfu: agree

PR #235, opened 2020-02-23
make datacenter non required on rack filter

juliasaveliff:


clairecfu:
just pulled this into rack-fixes branch to confirm that it works and it does!

PR #239, opened 2020-02-23
Reporting updates

brendancheng20:
- Now supporting global and per-datacenter reports


rackcity/views/report_views.py
```diff
@@ -74,3 +74,75 @@ def report_rack_usage(request):
...
+@api_view(['GET'])
+@permission_classes([IsAuthenticated])
+def rack_report_datacenter(request, id):
+    """
+    Get summary of rack usage for ONE DATACENTER: the percentage of rackspace free versus used,
+    allocated per vendor, allocated per model, and allocated per owner.
+    """
+    total_num_rack_slots = 0
+    datacenter_id = id
+    racks = Rack.objects.filter(datacenter=datacenter_id)
```
juliasaveliff: it would be nice if we could make a helper method that takes in racks list and does the rest since the 2 endpoints are identical after this line just so we can keep the number of lines down ;))))

also this is probably a bug I wrote since I did the backend of report for ev1 but there is a row for "(no owner)" and a row for empty owner "" in owner allocation table

src/components/report/report.scss
```diff
@@ -33,3 +33,7 @@
 .row {
   display: flex;
 }
+
+.report-all {
```
juliasaveliff: not sure if this is a css thing but in terms of usability i think the font size is a lot smaller on the report page now?

juliasaveliff: not a huge deal but idk if we want them trying to zoom in during eval

PR #240, opened 2020-02-23
Beautify error messages PART 1

juliasaveliff:
These changes are getting increasingly massive so I'm splitting up into multiple PRs. Some of the changes look deceptively as though I have rewritten the entire endpoint. So sorry for the scary line numbers but I did NOT change all the implementation - it's because all updated endpoints are now consistent in failing early with detailed error messages. If any of the endpoints look like all the lines have changed, it's probably from moving around the "if ... else ..." and "try ... except ... else" blocks where we were previously not failing early

This PR updates error messages for the following views:
[x] User
[x] Rack
[x] Model (but NOT bulk actions) 

Errors to conquer in future PRs: 
[ ] Model bulk actions
[ ] Asset 
[ ] Asset bulk actions
[ ] Make serializer errors more detailed (show if uniqueness, invalid value issue, etc...)
[ ] Datacenter
[ ] PDU


rackcity/api/objects/rack_range.py
```diff
@@ -62,6 +62,10 @@ def get_num_racks_in_range(self):
             num_col = 1
         return num_row * num_col
 
+    def get_range_as_string(self):
+        return "{" + self.get_row_range_as_string() + "}{" + \
+            self.get_number_range_as_string() + "}"
+
```
juliasaveliff: So we can display rack ranges like `{A-B}{1-2}` easily in error messages

rackcity/views/it_model_views.py
```diff
@@ -50,25 +54,46 @@ def model_add(request):
...
     if 'id' in data:
-        failure_message = failure_message + "Don't include id when adding a model. "
+        return JsonResponse(
+            {
+                "failure_message":
+                    Status.CREATE_ERROR.value + GenericFailure.UNKNOWN.value,
+                "errors": "Don't include 'id' when creating a model"
+            },
+            status=HTTPStatus.BAD_REQUEST
+        )
```
juliasaveliff: Updating all endpoints to fail early with explicit error messages based on previous convos

rackcity/views/it_model_views.py
```diff
@@ -50,25 +54,46 @@ def model_add(request):
...
-    return JsonResponse({
-        "failure_message": failure_message
-    },
-        status=HTTPStatus.NOT_ACCEPTABLE
+        return JsonResponse(
+            {
+                "failure_message":
+                    Status.CREATE_ERROR.value +
+                    GenericFailure.INVALID_DATA.value,
+                "errors": str(serializer.errors)
```
juliasaveliff: This means we lose some detail (uniqueness validation, invalid value validation, etc) because we're not returning the serializer's json gobbly gook answer. To get the specifics of what went wrong in the serialization, we could parse the serializer errors, but saving that improvement for another PR / time permitting

rackcity/views/it_model_views.py
```diff
@@ -79,38 +104,60 @@ def model_modify(request):
...
+                    "Model" + GenericFailure.DOES_NOT_EXIST.value,
+                "errors": "No existing model with id="+str(id)
+            },
+            status=HTTPStatus.BAD_REQUEST
+        )
+    try:
+        validate_model_height_change(data, existing_model)
+    except LocationException as error:
+        location_failure = "Height change of model causes conflicts. " + \
+            str(error)
```
juliasaveliff: str(error) is ok for failure message here because they are nice looking errors that Ben wrote up in LocationException

PR #245, opened 2020-02-24
Log fixes and improvements

juliasaveliff:
remember when ben sat on an epic log?

### DIFF
--- a/rackcity/utils/log_utils.py
+++ b/rackcity/utils/log_utils.py
@@ -63,6 +63,7 @@ def log_action(user, related_element, action):
         datetime_to_string(date),
         element_type,
         element_name + ":",
+        "user",
         user.username,
         action.value,
         element_type,
@@ -87,6 +88,7 @@ def log_delete(user, element_type, element_name):
         datetime_to_string(date),
         element_type.value,
         element_name + ":",
+        "user",
         user.username,
         Action.DELETE.value,
         element_type.value,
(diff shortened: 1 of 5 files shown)


rackcity/utils/log_utils.py
```diff
@@ -163,29 +167,19 @@ def log_power_action(user, power_action, related_asset):
     log.save()
 
 
-def log_network_action(user, action, asset_0, asset_1):
+def log_network_action(user, related_asset):
```
juliasaveliff: there was no clear point in the creation/deletion of network connections where it made sense to call this log. so now network connection actions will be logged only via the asset they were done from

src/components/logs/logs.scss
```diff
@@ -34,4 +34,7 @@ form {
   -moz-box-sizing: content-box;
   -webkit-box-sizing: content-box;
   box-sizing: content-box;
+}
+.log-block {
+  overflow: scroll;
```
clairecfu: wooo

rackcity/utils/log_utils.py
clairecfu: do we need to have logging for bulk import? / will this support network bulk import?

juliasaveliff: going to go back and perfect the bulk import logging and errors in one swoop later!

PR #248, opened 2020-02-24
Beautify error messages PART 3

juliasaveliff:
Previous PR
- user views
- rack views
- model (non-bulk) views
- datacenter views

**This PR**
- **asset (non-bulk) views**

Future PR
- model bulk views
- asset bulk views
- PDU views

### DIFF
--- a/rackcity/utils/errors_utils.py
+++ b/rackcity/utils/errors_utils.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ValidationError
 from enum import Enum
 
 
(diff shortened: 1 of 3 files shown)


rackcity/utils/errors_utils.py
```diff
@@ -85,3 +86,13 @@ def parse_serializer_errors(errors):
...
+
+
+def parse_validation_error(error):
+    failure_detail = ""
+    if isinstance(error, ValidationError):
+        for err in error:
+            failure_detail += err
+    else:
+        failure_detail = GenericFailure.UNKNOWN.value
+    return failure_detail
```
juliasaveliff: if model.save() or serializer.save() fails due to one of our custom ValidationError's we can show what that is

PR #250, opened 2020-02-24
Check if asset can move to new datacenter

juliasaveliff:
Modification of asset's datacenter (rack) fails if the asset has existing power connections or network connections


rackcity/api/serializers/asset_serializers.py
```diff
@@ -4,7 +4,8 @@
 from rackcity.models import Asset, PowerPort, NetworkPort
 from .it_model_serializers import ITModelSerializer
 from .rack_serializers import RackSerializer
-import copy 
+import copy
+
 
 class AssetSerializer(serializers.ModelSerializer):
```
juliasaveliff: only auto linting in this file

PR #251, opened 2020-02-24
Asset form cleanup

clairecfu:
- [x] display warnings on asset modification/creation -> a WARN toast, with a button that opens up the edit form again 
- [x] allows users to clear/delete network connections and power connections 
- [x] general asset form cleanup
- [x] no more hard coded form labels


src/utils/utils.tsx
```diff
@@ -54,11 +54,34 @@ export interface RackRangeFields {
...
+  model__model_number: "Model Number",
+  rack: "Rack",
+  rack__datacenter__name: "Datacenter",
+  rack_position: "Rack Position",
+  owner: "Owner"
+};
+
+export enum AssetFormLabels {
+  asset_number = "Asset Number*",
+  hostname = "Hostname*",
```
juliasaveliff: does * indicate required? might be confusing since these are not required for user to input

PR #252, opened 2020-02-24
Suggested asset number 

clairecfu:
- [x] full stack feature, autofill suggested asset number on asset creation


rackcity/models/asset.py
```diff
@@ -7,6 +7,13 @@
 from .rack import Rack
 
 
+def get_asset_number():
```
juliasaveliff: I would maybe rename something like "get_next_available_asset_number", sort of a long name but makes it more obvious whats goin on??

clairecfu: coool

PR #260, opened 2020-02-25
Locking model modifications 

clairecfu:


rackcity/views/it_model_views.py
```diff
@@ -162,6 +170,22 @@ def model_modify(request):
     )
 
 
+def validate_model_change(new_model_data, existing_model): 
+    if "network_ports" not in new_model_data:
+        return
+    if "num_power_ports" not in new_model_data:
+        return
+    assets = Asset.objects.filter(model=existing_model.id)
+    if assets and len(assets) > 0:
```
juliasaveliff: you don't need to check `if assets` here because it will return an empty queryset if there are no assets matching the filter, so `len` would not throw an error

src/forms/forms.scss
```diff
@@ -7,7 +7,7 @@
 }
 .login-form {
   display: inline-block;
-  padding: 10pt;
+  padding: 20pt;
```
juliasaveliff: this was changed to 10pt for my PR to make login nicer

PR #275, opened 2020-02-25
Ev2 Testing bug fixes 1

clairecfu:
- [x] property names on asset and model detail view pages
- [x] model form does not break when negative values are entered for no network ports 
- [x] showing datacenter on asset detail view page

### DIFF
--- a/src/forms/modelForm.tsx
+++ b/src/forms/modelForm.tsx
@@ -124,6 +124,10 @@ class ModelForm extends React.Component<ModelFormProps, ModelFormState> {
       let num_network_ports = field["num_network_ports"];
       console.log(num_network_ports, network_ports);
       let index = network_ports.length;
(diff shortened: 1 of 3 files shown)


src/forms/modelForm.tsx
```diff
@@ -124,6 +124,10 @@ class ModelForm extends React.Component<ModelFormProps, ModelFormState> {
       let num_network_ports = field["num_network_ports"];
       console.log(num_network_ports, network_ports);
       let index = network_ports.length;
+      if (!isNaN(num_network_ports) && num_network_ports < 0) {
+        num_network_ports = 0;
+      }
```
juliasaveliff: will this result in the `num_network_ports` field being sometimes `0` and sometimes `null` (in the DB and also in the table wherever we display it? not sure if that is ideal, but not a dealbreaker, just something to track

PR #277, opened 2020-02-25
Changing power button names and (required) fields to (*)

brendancheng20:


src/forms/datacenterForm.tsx
```diff
@@ -1,142 +1,140 @@
...
+    return (
+      <div className={Classes.DARK + " login-container"}>
+        {this.state.errors.map((err: string) => {
+          return <Callout intent={Intent.DANGER}>{err}</Callout>;
+        })}
+        <form
+          onSubmit={this.handleSubmit}
+          className="create-form bp3-form-group"
+        >
+          <FormGroup label="Name (*)" inline={false}>
```
juliasaveliff: for consistency with the other forms, this would be "Name*" and "Abbreviation* (up to 6 characters)" I know that's picky but this was a picky change to begin with muahaa

brendancheng20: Done!

PR #279, opened 2020-02-25
showing errors on inline modify

clairecfu:
- [x] fixed for assets, models, and datacenters

### DIFF
--- a/src/components/elementView/elementTable.tsx
+++ b/src/components/elementView/elementTable.tsx
@@ -640,12 +640,11 @@ class ElementTable extends React.Component<
   };
   handleEditFormSubmit = (values: ElementObjectType, headers: any) => {
     if (isModelObject(values)) {
-      modifyModel(values, headers).then(res => {
+      return modifyModel(values, headers).then(res => {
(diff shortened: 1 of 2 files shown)


src/forms/formPopup.tsx
```diff
@@ -24,6 +24,7 @@ interface FormPopupProps {
   currDatacenter?: DatacenterObject;
   initialValues?: ElementObjectType;
   elementName: ElementType;
+  loading?: boolean;
```
clairecfu: this is accidentally a change meant for another branch, but not breaking since its optional

src/components/elementView/elementTable.tsx
```diff
@@ -640,20 +640,19 @@ class ElementTable extends React.Component<
   };
   handleEditFormSubmit = (values: ElementObjectType, headers: any) => {
     if (isModelObject(values)) {
-      modifyModel(values, headers).then(res => {
+      return modifyModel(values, headers).then(res => {
```
clairecfu: was not returning the response! so nothing to catch

PR #280, opened 2020-02-25
Set owner to null when an asset's owner is deleted from users

juliasaveliff:


rackcity/models/asset.py
```diff
@@ -74,8 +74,7 @@ class Meta:
     def save(self, *args, **kwargs):
         try:
             validate_hostname(self.hostname)
-            if self.owner is not None:
```
juliasaveliff: Moved this to inside `validate_owner` to be consistent with validate_hostname

rackcity/views/user_views.py
```diff
@@ -239,6 +240,10 @@ def user_delete(request):
             status=HTTPStatus.BAD_REQUEST,
         )
     else:
+        deleted_user_assets = Asset.objects.filter(owner=username)
+        for asset in deleted_user_assets:
+            asset.owner = None
+            asset.save()
```
juliasaveliff: after deleting the user, iterate over any assets who had the deleted user's username, set its owner to null

PR #282, opened 2020-02-26
fixed bug with modifying network or power interface of a model via bulk

benauriemma:


rackcity/views/it_model_views.py
```diff
@@ -177,17 +177,31 @@ def model_modify(request):
...
     assets = Asset.objects.filter(model=existing_model.id)
     if len(assets) > 0:
-        if set(new_model_data["network_ports"]) != set(existing_model.network_ports):
-            raise ModelModificationException("Unable to modify network ports.")
-        if int(new_model_data["num_power_ports"]) != existing_model.num_power_ports:
+        if (
+            set(new_model_data["network_ports"])
+            != set(existing_model.network_ports)
+        ):
+            raise ModelModificationException(
```
clairecfu: in that case we should also add `"network_ports" in new_model_data ` to this condition

rackcity/views/it_model_views.py
```diff
@@ -177,17 +177,31 @@ def model_modify(request):
...
+        if (
+            (
+                not new_model_data["num_power_ports"]
+                and existing_model.num_power_ports
+            )
+            or (
+                int(new_model_data["num_power_ports"])
+                != existing_model.num_power_ports
+            )
+        ):
```
clairecfu: and `"num_power_ports" in new_model_data` right

PR #291, opened 2020-02-26
DONE WITH ERROR MESSAGES TONIGHT!

juliasaveliff:


rackcity/utils/errors_utils.py
```diff
@@ -15,11 +16,14 @@ class Status(Enum):
 
 
 class GenericFailure(Enum):
-    UNKNOWN = "Unknown internal error."
+    INTERNAL = "Internal error."
```
juliasaveliff: this is used SPARINGLY only when `id` is not specified in a request or it is specified when it shouldn't be

rackcity/utils/errors_utils.py
```diff
@@ -15,11 +16,14 @@ class Status(Enum):
...
     DOES_NOT_EXIST = " does not exist."
     FILTER = "Invalid filter applied."
     SORT = "Invalid sort applied."
     INVALID_DATA = "Required values are missing or invalid."
+    ON_SAVE = " failed to save. Please ensure all input values are valid."
+    ON_DELETE = " failed to delete due to internal error."
+    PAGE_ERROR = "Failed to load this page of data. Please try again."
```
juliasaveliff: made more descriptive errors to replace "unknown internal error"

rackcity/utils/errors_utils.py
```diff
@@ -106,11 +110,14 @@ def parse_serializer_errors(errors):
...
     failure_detail = ""
     if isinstance(error, ValidationError):
         for err in error:
             failure_detail += err
+    elif isinstance(error, IntegrityError):
+        for err in error.args:
+            failure_detail += err
     else:
-        failure_detail = GenericFailure.UNKNOWN.value
+        failure_detail = object_name + GenericFailure.ON_SAVE.value
```
juliasaveliff: only used when this is NOT a ValidationError (custom validation errors we have created) or IntegrityError (error thrown from the depths of the DB on save). no idea when that would even happen

PR #308, opened 2020-03-02
biggg routing changes 

clairecfu:
- [x] new routing system so that dashboard maintains state (does not unmount and remount)
- [x] good bye hard coded routes (use enum in utils.tsx)
- [x] adding required isauthenticated for i-am-admin
- [x] showing tabs on detailview pages 
- [x] css changes


rackcity/views/user_views.py
```diff
@@ -292,6 +292,7 @@ def who_am_i(request):
 
 
 @api_view(['GET'])
+@permission_classes([IsAuthenticated])
```
clairecfu: @juliasaveliff as discussed

PR #309, opened 2020-03-06
Claire change planner

clairecfu:
- [x] creating asset change planner for @benauriemma  to refactor hehe


rackcity/models/asset_change_planner.py
```diff
@@ -0,0 +1,79 @@
...
+        validators=[validate_hostname],
+        null=True,
+        blank=True,
+    )
+    rack_position = models.PositiveIntegerField()
+    model = models.ForeignKey(
+        ITModel,
+        on_delete=models.CASCADE,
+        verbose_name="related model",
+    )
```
benauriemma: FYSA, `on_delete=models.CASCADE` means that if a model is deleted, any AssetCP with that model will also be deleted without warning. I'm not sure what the desired behavior is there but if it's left as `CASCADE` then you'll likely want to add something to the check we do for existing assets when deleting a model

rackcity/models/asset_change_planner.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class AssetCP(models.Model):
+    related_asset = models.ForeignKey(
+        Asset,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+    )
+    ## TODO: related_decomissioned_asset
```
benauriemma: What will this field be?

rackcity/models/change_plan.py
```diff
@@ -0,0 +1,22 @@
...
+from django.db import models
+from django.contrib.auth.models import User
+
+
+class ChangePlan(models.Model):
+    name = models.CharField(max_length=150)
+    execution_time = models.DateTimeField(blank=True, null=True)
+    owner = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE
```
benauriemma: Same comment here about `on_delete=models.CASCADE` - if a user is deleted, their change plan will disappear without warning. I'm not sure that's the desired behavior.

rackcity/models/change_plan.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+
+    class Meta:
+        ordering = ['name']
+        constraints = [
+            models.UniqueConstraint(
+                fields=['name', 'owner'],
```
benauriemma: This says that two different users can make change plans with the same name - is that what we want?

rackcity/models/asset_change_planner.py
clairecfu: this will be the same idea as related_asset but will point to a decomissionedAsset object once it exists

rackcity/models/change_plan.py
clairecfu: this is the expected behavior we settled on, since only users can see their own change plans, also clarified on piazza via this post 
https://piazza.com/class/k4u27qnccr45oo?cid=139

rackcity/models/change_plan.py
clairecfu: yeah - same idea as above comment, only individuals can see their own change plans

rackcity/models/asset_change_planner.py
clairecfu: changing this to SET_NULL, for both model and rack.

rackcity/models/asset_change_planner.py
```diff
@@ -44,12 +44,14 @@ class AssetCP(models.Model):
     rack_position = models.PositiveIntegerField()
     model = models.ForeignKey(
         ITModel,
-        on_delete=models.CASCADE,
+        on_delete=models.SET_NULL,
+        null=True,
         verbose_name="related model",
     )
```
clairecfu: expected behavior if you the model or rack in the real world is deleted: the change planner will show a conflict saying "associated model was deleted"... this would also result in invalid network ports/power port connections, not sure how we want to handle this exactly, but will discuss with @juliasaveliff

PR #311, opened 2020-03-09
asset inheritance

benauriemma:


rackcity/models/asset.py
```diff
@@ -32,7 +32,11 @@ def validate_owner(value):
         )
 
 
-class Asset(models.Model):
+class AbstractAsset(models.Model):
+    id = models.AutoField(primary_key=True)
```
benauriemma: Important thing to note here is that while this is called `AbstractAsset`, it is actually NOT an abstract model.

Django allows you to make an abstract model by setting `abstract = True` in the Meta class.  With `abstract = True`, there is no table created for the model, so you could use it as a parent and allow children to inherit from it, but you would not be able to instantiate the abstract parent.  This would be preferred for AbstractAsset.

The problem is that if AbstractAsset is abstract, the id will not be shared between the Asset and AssetCP.  So, in this implementation we do not set `abstract = True` on AbstractAsset.  This means that the id is shared (desired behavior), but it means we actually have a table of AbstractAssets which we'll never interact with (unfortunate side effect).

PR #312, opened 2020-03-16
User groups & permissions

juliasaveliff:
- create groups
- create permissions
- endpoints to set and get user groups

todo in future
- add datacenter-specific asset permissions (as field on user object)
- create and add wrappers to existing endpoints
- front end changes


.vscode/settings.json
```diff
@@ -3,14 +3,13 @@
...
   "editor.formatOnSave": true,
-
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-
   "python.linting.enabled": true,
   "python.linting.lintOnSave": true,
+  "python.linting.mypyEnabled": true,
```
juliasaveliff: allows (but does not enforce) static typing in python functions

rackcity/views/user_views.py
```diff
@@ -442,3 +447,108 @@ def usernames(request):
         {"usernames": usernames},
         status=HTTPStatus.OK,
     )
+
+
+@api_view(['POST'])
+@permission_classes([IsAdminUser])
+def user_set_groups(request):
```
juliasaveliff: endpoint to set groups for a user (add or remove them)

rackcity/views/user_views.py
```diff
@@ -442,3 +447,108 @@ def usernames(request):
...
+            "User's groups were not changed."
+    return JsonResponse(
+        {"success_message": Status.SUCCESS.value + success_message},
+        status=HTTPStatus.OK,
+    )
+
+
+@api_view(['POST'])
+@permission_classes([IsAuthenticated])
+def user_get_groups(request):
```
juliasaveliff: endpoint to get the groups that a particular user belongs to (returned as a list of strings of group names)

rackcity/views/user_views.py
```diff
@@ -442,3 +447,108 @@ def usernames(request):
...
+        group_list.append(group.name)
+    return JsonResponse(
+        {"user_groups": group_list},
+        status=HTTPStatus.OK,
+    )
+
+
+@api_view(['GET'])
+@permission_classes([IsAuthenticated])
+def all_user_groups(request):
```
juliasaveliff: endpoint to get list of all groups (group names as strings)

.vscode/settings.json
benauriemma: Very lyft of you

rackcity/views/user_views.py
```diff
@@ -442,3 +447,108 @@ def usernames(request):
...
+                if added:
+                    groups_added.append(group_key)
+            else:
+                removed = remove_user_from_group(user, group)
+                if removed:
+                    groups_removed.append(group_key)
+    success_message = ""
+    if len(groups_added) > 0:
+        success_message += \
+            ("User added to groups: " + ", ".join(groups_added) + ". ")
```
benauriemma: Do you think it'll be concerning if you add a user to Models and Assets, but they were already in Assets, so the success message only says they were added to Models?  What if the success message also said something like "User was already in groups: Assets"

benauriemma: This may depend on what the workflow is like on the frontend - like if it's a multi-select you may be sending the redundant ones every time so maybe you don't want that in the success message.  Also possible you'd never have the case where you thought you added them but they were already in it

juliasaveliff: I like that; I'll modify it

juliasaveliff: can't believe I used a semicolon

rackcity/views/user_views.py
```diff
@@ -442,3 +447,115 @@ def usernames(request):
...
+            ("User removed from group(s): " + ", ".join(groups_removed) + ". ")
+    if len(groups_added) == 0 and len(groups_removed) == 0:
+        success_message += \
+            "User's groups were not changed. "
+    if len(current_groups) > 0:
+        success_message += \
+            "User is now in group(s): " + ", ".join(current_groups) + "."
+    else:
+        success_message += \
+            "User is now in no groups."
```
juliasaveliff: updated success message

rackcity/views/user_views.py
benauriemma: A period would have been fine

PR #318, opened 2020-03-24
Change Plan Models

clairecfu:
- [x] NetworkPortCP
- [x] PowerPortCP
- [x] PDUPortCP

assetCP still unfinished, waiting on decomissioned asset


clairecfu:
I haven't fully tested all the new models (since have no endpoints yet) but I did my best to test to make sure the refactoring didnt break any existing functiohnality

clairecfu:
also in this PR i realized that the front end was not using a lot of success and error messages, and instead had hard coded messages so i tried to fix most of it

rackcity/models/network_port.py
```diff
@@ -84,7 +97,22 @@ class Meta:
...
+class NetworkPortCP(AbstractNetworkPort):
+    asset = models.ForeignKey(
+        AssetCP,
+        on_delete=models.CASCADE,
+        verbose_name="asset",
+    )
+    change_plan = models.ForeignKey(
+        ChangePlan,
+        on_delete=models.CASCADE,
+    )
```
juliasaveliff: is it redundant to store change plan here? shouldn't it always be the same change plan as `asset.change_plan`?

clairecfu: uniqueness constraint? but not sure

rackcity/admin.py
```diff
@@ -1,6 +1,5 @@
 from django.contrib import admin
 from .models import (
-    AbstractAsset,
```
benauriemma: Don't delete pls 🤪

rackcity/models/asset.py
```diff
@@ -32,10 +33,32 @@ def validate_owner(value):
         )
 
 
-class AbstractAsset(models.Model):
+class AbstractID(models.Model):
     id = models.AutoField(primary_key=True)
 
 
+class AbstractAsset(AbstractID):
```
benauriemma: I would name these differently - specifically I think AbstractID is unclear bc it doesn't have the word Asset

rackcity/models/asset.py
```diff
@@ -114,3 +122,87 @@ def add_power_ports(self):
                         port_name=port_name,
                     )
                     power_port.save()
+
+
+class AssetCP(AbstractAsset):
```
benauriemma: Why'd you move this to the same file?

rackcity/models/pdu_port.py
```diff
@@ -24,9 +25,29 @@ class PDUPort(models.Model):
...
+class PDUPort(AbstractPDUPort):
     class Meta:
         constraints = [
             models.UniqueConstraint(
                 fields=['rack', 'left_right', 'port_number'],
                 name='unique port per pdu'),
         ]
+
+
+class PDUPortCP(AbstractPDUPort):
```
benauriemma: I think I'm more fine with the port CPs being in the same file bc they're smaller, they don't have a decommissioned counterpart, and they don't really have a ton of logic

src/components/elementView/elementTab.tsx
```diff
@@ -236,7 +236,7 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
       console.log("success");
       this.handleDataUpdate(true);
       this.handleClose();
-      this.addSuccessToast("Successfully created model!");
+      this.addSuccessToast(res.data.success_message);
```
benauriemma: Did you confirm that there's a success_message returned in all the places you changed this?  I'm not sure every backend endpoint returns one

rackcity/admin.py
clairecfu: oops - this should be AssetID! Django doesn't let you have true abstract classes in admin.py

rackcity/models/asset.py
clairecfu: changing to AssetID

rackcity/models/asset.py
clairecfu: see comment below

rackcity/models/pdu_port.py
clairecfu: yeah, I thought it was confusing to have the Abstract class and one of its subclasses in one file, and another one of its subclasses in a diff file. So either theyre all in separate files or theyre all in 1 file. I thought for sake of simplicity I'd combine them all into one file so we dont have a ton of files.

src/components/elementView/elementTab.tsx
clairecfu: yeah im 99% sure  -> these changes were made when julia worked on pretty error messages (which also included success messages). this is actually bad that we were hard coding this info because we were losing on information being sent from backend.

PR #320, opened 2020-03-24
Datacenter asset permissions & power permissions

juliasaveliff:
- RackCityPermission model to store per-datacenter user permissions
- Decorators for more complicated asset endpoint permissions & power endpoint permissions
- Set per-datacenter user permissions added to existing endpoint

as mentioned on slack, there are some TODOs in this PR that I would like to save for the next PR because this one is getting beefy


rackcity/api/urls.py
```diff
@@ -52,8 +52,8 @@
     path('users/pages', views.user_page_count),
     path('users/grant-admin', views.user_grant_admin),
     path('users/revoke-admin', views.user_revoke_admin),
-    path('users/groups/get', views.user_get_groups),
-    path('users/groups/set', views.user_set_groups),
+    path('users/permissions/get', views.user_get_groups),
+    path('users/permissions/set', views.user_set_groups),
```
juliasaveliff: changed these endpoint names because the same endpoint is used to set asset datacenter-level permissions

rackcity/views/pdu_views.py
```diff
@@ -15,23 +15,25 @@
...
+
 pdu_url = 'http://hyposoft-mgt.colab.duke.edu:8005/'
 # Need to specify rack + side in request, e.g. for A1 left, use A01L
 get_pdu = 'pdu.php?pdu=hpdu-rtp1-'
 toggle_pdu = 'power.php'
 
 
 @api_view(['GET'])
-@permission_classes([IsAuthenticated])
+@power_permission_required()
```
benauriemma: Is power permission required to see the power status of an asset? I didn't think it was

juliasaveliff: oh you're right I was wrong. fixing!

PR #322, opened 2020-03-25
Helper method to get assets for a branch

juliasaveliff:


rackcity/utils/change_planner_utils.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import Q
+from rackcity.models import Asset, AssetCP
+
+
+def get_assets_for_branch(change_plan=None):
```
clairecfu: can we rename this to get_assets_for_cp for consistency

PR #323, opened 2020-03-25
get many for decommissioned

benauriemma:
- [x] filtering works, including nested fields
- [x] sorting works, including nested fields
- [ ] no date filter


rackcity/views/decommission_views.py
```diff
@@ -87,3 +92,100 @@ def decommission_asset(request):
...
+
+
+@api_view(['POST'])
+@permission_classes([IsAuthenticated])
+def decommissioned_asset_many(request):
+    """
+    List many decommissioned assets. If page is not specified as a query parameter, all
+    assets are returned. If page is specified as a query parameter, page
+    size must also be specified, and a page of assets will be returned.
+    """
```
benauriemma: The way this is written definitely takes on some technical debt - this view is almost exactly the same as `asset_many`, and shares a lot of code for paging, sorting, and filtering with some other get-many endpoints.  I personally don't think it's worth refactoring as I think that would cost more time now than the repeated code will cost us next evolution, but I'm open to dissent

benauriemma: Sike Julia just refactored this

benauriemma: Tagging WIP until that's merged and I can change this to reflect it

PR #324, opened 2020-03-26
Get many assets updated for CP

juliasaveliff:
To get many assets on a change planner, filtering works! but sorting and paging will be pushed in another PR


rackcity/api/serializers/asset_serializers.py
```diff
@@ -193,11 +221,32 @@ def normalize_bulk_asset_data(bulk_asset_data):
     return bulk_asset_data
 
 
-def serialize_power_connections(asset):
-    try:
-        ports = PowerPort.objects.filter(asset=asset.id)
-    except ObjectDoesNotExist:
-        return
+def serialize_mac_addresses(network_port_model, asset):
+    ports = network_port_model.objects.filter(asset=asset.id)
```
juliasaveliff: Removed the try/except around this because `.filter` does not throw doesnotexist exceptions--it will just return an empty queryset if no existing rows match the filter

rackcity/utils/change_planner_utils.py
```diff
@@ -18,3 +25,32 @@ def get_assets_for_cp(change_plan=None):
...
+        return filter_failure_response
+    asset_serializer = RecursiveAssetSerializer(
+        filtered_assets,
+        many=True,
+    )
+    assetCP_serializer = RecursiveAssetCPSerializer(
+        filtered_assetsCP,
+        many=True,
+    )
+    all_assets = asset_serializer.data + assetCP_serializer.data
```
juliasaveliff: Need to add sorting and paging at this point in another PR

rackcity/utils/query_utils.py
```diff
@@ -173,9 +173,9 @@ def get_filtered_query(object_query, data, or_filters=False):
             )
         )
     if or_filters and len(filter_args) > 0:
-        apply_filters_or(object_query, filter_args)
+        object_query = apply_filters_or(object_query, filter_args)
     else:
-        apply_filters_and(object_query, filter_args)
+        object_query = apply_filters_and(object_query, filter_args)
```
juliasaveliff: bug fixes

rackcity/api/serializers/asset_serializers.py
```diff
@@ -171,6 +156,49 @@ def power_port_connection(self, asset, port_number):
...
+            'hostname',
+            'model',
+            'rack',
+            'rack_position',
+            'owner',
+            'comment',
+            'mac_addresses',
+            'network_connections',
+            'network_graph',
+            'power_connections',
```
clairecfu: could be useful to add `change_plan` as a field to the serializer and display this on the front end - but can be added later if we decide we need it

rackcity/views/asset_views.py
```diff
@@ -78,6 +79,26 @@ def asset_many(request):
...
+            return JsonResponse(
+                {
+                    "failure_message":
+                        Status.ERROR.value +
+                        "Change Plan" + GenericFailure.DOES_NOT_EXIST.value,
+                    "errors":
+                        "No existing change plan with id="+str(change_plan_id)
+                },
+                status=HTTPStatus.BAD_REQUEST
+            )
```
clairecfu: I have a helper function that does basically this ill do the refactor on my branch!

rackcity/views/asset_views.py
```diff
@@ -78,6 +79,26 @@ def asset_many(request):
     assets are returned. If page is specified as a query parameter, page
     size must also be specified, and a page of assets will be returned.
     """
+    if 'change_plan' in request.data:
+        change_plan_id = request.data['change_plan']
+        try:
```
clairecfu: not sure if you have strong thoughts about this, but for create/modify endpoints I have been assuming change_plan would be a query param, not in the request body

juliasaveliff: I don't feel strongly and I'd like them to match so I'll make it a query param here

rackcity/views/asset_views.py
juliasaveliff: we <3 refactoring

juliasaveliff: we </3 rewriting

rackcity/api/serializers/asset_serializers.py
juliasaveliff: t'has been added

PR #326, opened 2020-03-26
CP detail view endpoint

juliasaveliff:


rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +178,117 @@ def get_page_count_response_for_cp(request, change_plan):
...
     page_count = math.ceil(count / page_size)
     return JsonResponse({"page_count": page_count})
+
+
+def get_changes_on_asset(asset, assetCP):
+    fields = [field.name for field in Asset._meta.fields]
+    changes = []
+    for field in fields:
+        if getattr(asset, field) != getattr(assetCP, field):
+            # TODO will need to serialize field if it's another model
```
clairecfu: so this would be rack/model... what about network and power connections? since those aren't fields on asset. You could you reverse relations to get all the network connections and power connections that Asset and AssetCP are forieign keys for and compare those? if you want to leave that in a separate PR im also good with that.

clairecfu: but we would need to do that in order to highlight changes to network /power connections

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +178,117 @@ def get_page_count_response_for_cp(request, change_plan):
...
+            })
+    return changes
+
+
+def get_cp_modification_conflicts(asset_cp):
+    conflicts = []
+    nonresolvable_message = "This conflict cannot be resolved; the change " + \
+        "needs to be removed from your change plan."
+    conflicting_asset_message_1 = "Due to more recent live changes, " + \
+        "your change planner modification to this asset's "
```
clairecfu: small nit but maybe instead of modifications, say `your change plan's version of this's assets: ` because a conflict could come from a created asset on cp.

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +178,117 @@ def get_page_count_response_for_cp(request, change_plan):
...
+def get_modifications_in_cp(change_plan):
+    assets_cp = AssetCP.objects.filter(change_plan=change_plan)
+    modifications = []
+    for asset_cp in assets_cp:
+        assetCP_data = RecursiveAssetCPSerializer(asset_cp).data
+        related_asset = asset_cp.related_asset
+        if asset_cp.is_decommissioned:
+            title = "Decommission asset"
+            if asset_cp.asset_number:
+                title += " " + str(asset_cp.asset_number)
```
clairecfu: if the asset_cp does not have an asset number, this is going to say Decomission asset None. Can we add a check for this and either leave it out or use its rack + rack location if the asset number doesn't exist?

clairecfu: "Decomission/Modify/Create" asset [asset_number] on [Rack] at [location]. This would be a helpful string if we use this for work orders as well. But add logic so that if asset_number is none it is just omitted and doesn't print as None

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +178,117 @@ def get_page_count_response_for_cp(request, change_plan):
...
+        else:
+            title = "Create asset"
+            if asset_cp.asset_number:
+                title += " " + str(asset_cp.asset_number)
+            asset_data = None
+        modifications.append({
+            "title": title,
+            "asset": asset_data,
+            "asset_cp": assetCP_data,
+            "conflicts": get_cp_modification_conflicts(asset_cp)
```
clairecfu: can you also include `change typ` and `changes` field? as documented on the notion

rackcity/utils/change_planner_utils.py
juliasaveliff: cool I like

rackcity/utils/change_planner_utils.py
juliasaveliff: I'm already checking `if asset_cp.asset_number:` here! But I like adding more detail about location I'll make that change

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +187,133 @@ def get_page_count_response_for_cp(request, change_plan):
...
     return JsonResponse({"page_count": page_count})
+
+
+def get_changes_on_asset(asset, assetCP):
+    fields = [field.name for field in Asset._meta.fields]
+    changes = []
+    for field in fields:
+        if getattr(asset, field) != getattr(assetCP, field):
+            changes.append(field)
+    # TODO check if network & power connections have changed
```
juliasaveliff: @clairecfu will do this in another PR

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +187,133 @@ def get_page_count_response_for_cp(request, change_plan):
...
+                title += get_location_detail(asset_cp)
+            asset_data = None
+        modifications.append({
+            "title": title,
+            "type": modification_type.value,
+            "asset": asset_data,
+            "asset_cp": assetCP_data,
+            "conflicts": get_cp_modification_conflicts(asset_cp),
+            "changes": changes,
+        })
```
juliasaveliff: @clairecfu updated the API object on notion slightly, check dat out

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +187,133 @@ def get_page_count_response_for_cp(request, change_plan):
...
+        if asset_cp.is_conflict:
+            conflicts.append({
+                "conflict_message":
+                    "Live changes have been made to this asset since your " +
+                    "latest change planner modification. This conflict " +
+                    "needs to be resolved. Please select which version " +
+                    "you would like to keep.",
+                "conflicting_asset": None,
+                "conflict_resolvable": True,
+            })
```
clairecfu: based on my comments on the slack thread, `if asset_cp.is_conflict && asset_cp.related_asset is None`: to check if the asset has been deleted on msater. and then mark this as a unresolvable conflict? conflict message `The asset being modified on this change plan has been deleted live. + non resolvable message`

rackcity/utils/change_planner_utils.py
```diff
@@ -178,3 +187,138 @@ def get_page_count_response_for_cp(request, change_plan):
...
+    nonresolvable_message = "This conflict cannot be resolved; the " + \
+        "related changes need to be removed from your change plan."
+    conflicting_asset_message_1 = "Due to more recent live changes, " + \
+        "your change plan version of this asset's "
+    conflicting_asset_message_2 = " now conflicts with a live asset. "
+    if asset_cp.is_conflict and asset_cp.related_asset is None:
+        conflicts.append({
+            "conflict_message":
+                "This asset has been deleted live. " + nonresolvable_message
+        })
```
juliasaveliff: @clairecfu added here

PR #330, opened 2020-03-27
Replace asset/power decorators to fix bugz

juliasaveliff:


rackcity/views/asset_views.py
```diff
@@ -165,6 +161,30 @@ def asset_add(request):
...
+            {
+                "failure_message":
+                    Status.AUTH_ERROR.value + AuthFailure.ASSET.value,
+                "errors":
+                    "User " + request.user.username +
+                    " does not have asset permission in datacenter id="
+                    + str(rack.datacenter.id)
+            },
+            status=HTTPStatus.UNAUTHORIZED
+        )
```
benauriemma: I don't love that you can fail in other ways BEFORE failing because you don't have permission but I see why you did it and don't think it's a big deal

rackcity/views/asset_views.py
```diff
@@ -413,8 +433,23 @@ def asset_modify(request):
             },
             status=HTTPStatus.BAD_REQUEST
         )
+    if not user_has_asset_permission(
```
benauriemma: Ditto

rackcity/views/asset_views.py
```diff
@@ -885,8 +951,7 @@ def asset_bulk_upload(request):
 
 
 @api_view(['POST'])
-@permission_required(PermissionPath.ASSET_WRITE.value, raise_exception=True)
-# TODO Check all assets for datacenter-level permissions
+@permission_classes([IsAuthenticated])
```
benauriemma: Do you not check asset permission on bulk approve anymore?

rackcity/views/asset_views.py
```diff
@@ -1116,8 +1195,7 @@ def network_bulk_upload(request):
...
```
benauriemma: Ditto. I guess this shouldn't ever be an issue in typical use

rackcity/views/asset_views.py
juliasaveliff: yeah per-datacenter kind of forces it to be like this, there's no use in checking for global asset permission at the beginning and need to make sure there's a valid rack in the request before you can check if the user has permission in that rack. ceste la vie

rackcity/views/asset_views.py
juliasaveliff: same as above, there's no way to check for permissions earlier because you need to be able to access the datacenter

rackcity/views/asset_views.py
juliasaveliff: it's checked below. for assets that are added, and within the location validation method for assets that are modified

juliasaveliff: oh sorry my b - it's not checked on asset / network connection bulk approve for the same reason we don't re-validate everything in approve modifications, any thing that's going to happen in approve should have been validated in bulk upload already

PR #336, opened 2020-03-28
Execute CP endpoint

juliasaveliff:
- Execute change planner endpoint
- Checks that there are no conflicts, then for each assetcp, updates (1) assets, (2) network connections, (3) power connections, and then optionally (4) decommissions


rackcity/utils/execute_change_planner_utils.py
```diff
@@ -0,0 +1,117 @@
...
+def get_updated_asset(asset_cp):
+    created = False
+    related_asset = asset_cp.related_asset
+    if related_asset:
+        updated_asset = related_asset
+    else:
+        updated_asset = Asset()
+        created = True
+    for field in Asset._meta.fields:
+        if field != 'id':
```
clairecfu: Not sure if you tested this and there were no issues, but you might need to also ignore this field: `field.name == "assetid_ptr"` because this is how it stores it's reference to AssetID, which actually stores the id field

clairecfu: i think when i left out `assetid_ptr` field i  accidentally overwrote the autofield,  resulting in two asset/assetCP with the same id.

rackcity/views/change_plan_views.py
```diff
@@ -254,3 +267,99 @@ def change_plan_detail(request, id):
...
+            {
+                "failure_message":
+                    Status.ERROR.value +
+                    "You do not have access to execute this change plan.",
+                "errors":
+                    "User " + request.user.username +
+                    "does not own change plan with id="+str(id)
+            },
+            status=HTTPStatus.BAD_REQUEST,
+        )
```
clairecfu: good to have this check in place, but generally this should never happen right? bc only a user can see their own chagne plans

rackcity/views/change_plan_views.py
```diff
@@ -254,3 +267,99 @@ def change_plan_detail(request, id):
         },
         status=HTTPStatus.OK
     )
+
+
+@api_view(['POST'])
+@permission_classes([IsAuthenticated])
+def change_plan_execute(request):
```
clairecfu: not sure if i missed this, but we also need to autogenerate Asset  number for AssetCP's without an asset number

rackcity/views/change_plan_views.py
juliasaveliff: correct

rackcity/views/change_plan_views.py
juliasaveliff: they are autogenerated on asset.save() for Asset

juliasaveliff: which is called in get_updated_asset() helper method

rackcity/utils/execute_change_planner_utils.py
juliasaveliff: cool thx 4 flagging i will update

rackcity/utils/execute_change_planner_utils.py
```diff
@@ -0,0 +1,117 @@
...
+def get_updated_asset(asset_cp):
+    created = False
+    related_asset = asset_cp.related_asset
+    if related_asset:
+        updated_asset = related_asset
+    else:
+        updated_asset = Asset()
+        created = True
+    for field in Asset._meta.fields:
+        if (field.name != 'id' and field.name != "assetid_ptr"):
```
juliasaveliff: updated @clairecfu

PR #337, opened 2020-03-28
User forms

brendancheng20:
This stuff is in bad need of a refactor, but i'll do that after ev3 probably


src/forms/userForm.tsx
```diff
@@ -0,0 +1,372 @@
...
+                  onChange={() => {
+                    this.setState({
+                      permissions: this.updateBooleanPermissions("audit_read")
+                    });
+                  }}
+                ></Checkbox>
+              </FormGroup>
+              <RadioGroup
+                inline={true}
+                label="Datacenter permissions"
```
juliasaveliff: can you label this "Asset Management" instead? that's how it's defined in the requirements

src/forms/userForm.tsx
```diff
@@ -0,0 +1,372 @@
...
+                  alignIndicator={Alignment.LEFT}
+                  disabled={this.state.permissions.admin}
+                  onChange={() => {
+                    this.setState({
+                      permissions: this.updateBooleanPermissions("audit_read")
+                    });
+                  }}
+                ></Checkbox>
+              </FormGroup>
+              <RadioGroup
```
juliasaveliff: ideally if administrator permission is checked, all Asset Management editing should be disabled too

is it possible to add another Check box next to Asset Management (currently "Datacenter permissions")? if I wanted to remove asset permission from a user, I don't think it's clear enough that you have to toggle to Per Datacenter and not select anything

src/forms/userForm.tsx
```diff
@@ -0,0 +1,372 @@
...
+                Submit
+              </Button>
+            </form>
+          </div>
+        )}
+      </div>
+    );
+  }
+
+  private renderDatacenterChecks() {
```
juliasaveliff: I think there's a bug - if I select Global, then select Per Datacenter, the boxes do not uncheck when you click on them

src/forms/userForm.tsx
juliasaveliff: another bug - selecting the Global radio button & submitting does not set asset_management correctly, it sends:
```
id: 8
model_management: false
asset_management: false
power_control: false
audit_read: true
admin: false
datacenter_permissions: []
```
(I had audit log checked so that one's right)

src/forms/userForm.tsx
```diff
@@ -0,0 +1,372 @@
...
+    var checks: Array<any>;
+    checks = [];
+    for (var i = 0; i < this.state.datacenters.length; i++) {
```
juliasaveliff: could we sort these alphabetically so it's easier for the user?? if that's a lot of extra work then don't worry about it

src/components/elementView/elementUtils.tsx
```diff
@@ -174,3 +175,9 @@ export const deleteUser = (
   const data = { id: user.id };
   return axios.post(API_ROOT + "api/users/delete", data, headers);
 };
+export const modifyUser = (
```
juliasaveliff: not sure where to put this comment bc idk where it's rendered - but is it possible to rename the inline button to "Edit permissions"? if not then maybe get rid of the text bc they can't edit the whole user so might be misleading?

brendancheng20: Yeah I had it as that initially but changed it to users bc "edit permissions" made a scroll bar popup and I didn't want to do the CSS to fix it. Fixed it now tho

src/forms/userForm.tsx
brendancheng20: Done!

src/forms/userForm.tsx
brendancheng20: Fixed

PR #346, opened 2020-03-29
decommissioned paging

benauriemma:


rackcity/views/decommission_views.py
```diff
@@ -287,3 +290,27 @@ def decommissioned_asset_many(request):
...
+    """
+    change_plan = None
+    if request.query_params.get("change_plan"):
+        (change_plan, response) = get_change_plan(
+            request.query_params.get("change_plan")
+        )
+        if response:
+            return response
+    if change_plan:
+        return get_page_count_response_for_cp(request, change_plan)
```
benauriemma: I think I copied this from before decommissioning was added to change plan - @clairecfu do you have SA?

PR #349, opened 2020-03-29
Permission bug fixes

juliasaveliff:
- Fix permissions on user registration endpoint
- Fix permissions on decommission asset endpoint

### DIFF
--- a/rackcity/views/user_views.py
+++ b/rackcity/views/user_views.py
@@ -1,10 +1,16 @@
 from django.contrib.auth.decorators import permission_required
+from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.contrib.auth.models import User, Group
 from django.core.exceptions import ObjectDoesNotExist
 from django.http import JsonResponse
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


rackcity/views/user_views.py
```diff
@@ -16,29 +22,25 @@
...
-)
+from rest_framework.permissions import IsAuthenticated
 
 
-class RegisterNameView(RegisterView):
+class RegisterNameView(PermissionRequiredMixin, RegisterView):
     serializer_class = RegisterNameSerializer
-    permission_classes = [IsAdminUser]
+    permission_required = PermissionPath.MODEL_WRITE.value
+    raise_exception = True
```
benauriemma: I'm not sure what this is but why does it require model write permission?

juliasaveliff: NICE CATCH i copy pasted the wrong thing. you need USER_WRITE permission for this tehe. FYSA this is a wrapper around django-rest-auth's user registration endpoint that we're using. it's the only class based view in the whole world that ive ever written

benauriemma: In the whole wide world ever?

PR #351, opened 2020-03-29
Show network/power connection info in detail view 

clairecfu:
- [x] also fixed a power availability bug

### DIFF
--- a/rackcity/views/pdu_views.py
+++ b/rackcity/views/pdu_views.py
@@ -332,6 +332,7 @@ def power_availability(request):
         if response:
             return response
         assets, assets_cp = get_assets_for_cp(change_plan.id)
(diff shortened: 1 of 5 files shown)


rackcity/views/pdu_views.py
```diff
@@ -332,6 +332,7 @@ def power_availability(request):
         if response:
             return response
         assets, assets_cp = get_assets_for_cp(change_plan.id)
+        assets = assets.filter(rack=rack.id)
```
juliasaveliff: will approve but want to flag this just in case / I'm not sure - why are you also not filtering assets_cp by rack?

clairecfu: oh thats a bug
fixigin

PR #362, opened 2020-03-29
made bulk upload not an admn route

benauriemma:


benauriemma:
This change makes it so that any authenticated user can access the bulk upload pages regardless of permission. I think that's fine because the button to navigate there is still greyed out correctly, and the backend doesn't allow you to upload anything if you don't have permission.  It would be nice to prevent people from accessing that route at all but I think that's lower priority

benauriemma:
(the bug it's fixing is that if you had model permission but aren't an admin, you couldn't access the bulk page, and the same for assets)

juliasaveliff:
I'm fine w this! I think the most important thing is that no un-authenticated users can ever reach interesting pages. if an authenticated user gets here, but can't do anything they're not allowed to do, i think that's fine

PR #364, opened 2020-03-29
Brendan bug fixes

brendancheng20:
- [ ] Blue barcode button -> white barcode button

- [ ] Removed scroll bar in user table

- [ ] Barcode button not clickable unless assets are selected for printing


src/components/userView/user.scss
```diff
@@ -1,5 +1,5 @@
 .user-view {
   margin: 20pt;
   margin-top: 0 !important;
-  overflow: auto;
+  overflow: hidden;
```
clairecfu: what was the issue with the scroll bar? we have been using overflow:auto everywhere

PR #365, opened 2020-03-29
fixed isChangePlan method

benauriemma:


src/utils/utils.tsx
```diff
@@ -302,7 +302,7 @@ export function isUserObject(obj: any): obj is UserInfoObject {
   return obj && obj.username;
 }
 export function isChangePlanObject(obj: any): obj is ChangePlan {
-  return obj && obj.name;
+  return obj && obj.name && !obj.abbreviation;
```
benauriemma: Previously datacenters tested positive 👀

clairecfu: OOF good CATCH ben

PR #366, opened 2020-03-29
Power port and network connections bug 

clairecfu:
fixed 2 separate bugs:

when Modifying asset with existing power connections -> modification goes through but power connections are not saved (removed ), and a warning message is returned 

When modifying an asset with existing network ports-> IntegrityError at /api/assets/modify
null value in column "rack_position" violates not-null constraint
DETAIL:  Failing row contains (f, null, null, null, null, null, 27, null, null, 134, 134, f, null, null, null, null).


rackcity/views/asset_views.py
```diff
@@ -412,10 +412,15 @@ def save_network_connections(asset_data, asset_id, change_plan=None):
                             change_plan=change_plan
                         )
                         # add destination asset to AssetCPTable
+                        print(destination_asset)
```
juliasaveliff: PRINT STATEMENTS haha.

rackcity/views/asset_views.py
```diff
@@ -412,10 +412,15 @@ def save_network_connections(asset_data, asset_id, change_plan=None):
...
+                        print(destination_asset._meta.fields)
                         for field in destination_asset._meta.fields:
-                            if field.name != 'id' and field.name == "assetid_ptr":
+                            
+                            if field.name != 'id' and field.name != "assetid_ptr":
```
juliasaveliff: it really be like that sometimes

rackcity/views/asset_views.py
```diff
@@ -506,16 +511,19 @@ def save_power_connections(asset_data, asset_id, change_plan=None):
                     left_right=power_connection_data['left_right'],
                     port_number=power_connection_data['port_number']
                 )
+                pdu_port = pdu_port_master
```
juliasaveliff: it seems like we prefer these johnsons to be "live" not "master" so maybe change to make consistent?

clairecfu: johnsons heh

rackcity/views/asset_views.py
clairecfu: yeah :/

PR #370, opened 2020-03-30
Form prepopulation bug fixes

brendancheng20:
- [x] Global asset management form prepopulation now works as expected

- [x] Giving per datacenter permissions no longer accidentally sends asset_management (global permission) as true

- [x] Unchecking admin does not leave all other boxes still checked unless they were checked before admin was checked

### DIFF
--- a/src/forms/userForm.tsx
+++ b/src/forms/userForm.tsx
@@ -27,7 +27,6 @@ import { FormTypes } from "./formUtils";
 import { API_ROOT } from "../utils/api-config";
 import { modifyUser } from "../components/elementView/elementUtils";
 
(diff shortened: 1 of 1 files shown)


src/forms/userForm.tsx
```diff
@@ -397,8 +397,9 @@ class UserForm extends React.Component<UserFormProps, UserFormState> {
           permissions: res.data
         });
         if (
-          res.data.datacenter_permissions.length ===
-          this.state.datacenters.length
+          res.data.asset_management
+          // res.data.datacenter_permissions.length ===
+          // this.state.datacenters.length
```
benauriemma: Can you delete the comment?

PR #371, opened 2020-03-30
Detect changed connections (for hightlighting fields)

juliasaveliff:
image

### DIFF
--- a/rackcity/utils/change_planner_utils.py
+++ b/rackcity/utils/change_planner_utils.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Q
 from django.db.models.signals import post_save, pre_delete
 from django.dispatch import receiver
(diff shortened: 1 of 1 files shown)


rackcity/utils/change_planner_utils.py
```diff
@@ -289,13 +297,65 @@ def get_page_count_response_for_cp(request, change_plan):
...
+        # Get Asset corresponding to this live connection
+        connected_port_live = network_ports.get(
+            port_name=network_port_cp.port_name
+        )
+        if connected_port_live:
+            destination_host_live = connected_port_live.asset
+        else:
+            destination_host_live = None
+        # Check for match
+        if destination_host_in_cp != destination_host_live:
```
clairecfu: my brain is tired. would this miss changes in destination port names? aka connected to same hostname but a different port?

juliasaveliff: good catch yeah I'll update

rackcity/utils/change_planner_utils.py
```diff
@@ -289,13 +298,69 @@ def get_page_count_response_for_cp(request, change_plan):
...
+                    real_connected_port_in_cp = NetworkPort.get(
+                        asset=related_asset,
+                        port_name=cp_connected_port_in_cp.port_name,
+                    )
+                except ObjectDoesNotExist:
+                    real_connected_port_in_cp = None
+        # Check for match
+        if connected_port_live != real_connected_port_in_cp:
+            return True
+    return False
```
juliasaveliff: this took some crazy brain power and drawings to get right I think it's right now it looks like it's working thank you and goodbye @clairecfu

benauriemma: I think you could have just checked the port name and hostname but go off queen

rackcity/utils/change_planner_utils.py
```diff
@@ -289,13 +298,69 @@ def get_page_count_response_for_cp(request, change_plan):
...
+        # Get PDUPort corresponding to this live connection
+        pdu_port_live = power_ports.get(port_name=power_port_cp.port_name)
+        # Check for match
+        if (pdu_port_cp is None and pdu_port_live is None):
+            continue
+        if (
+            (pdu_port_cp is None and pdu_port_live)
+            or (pdu_port_cp and pdu_port_live is None)
+        ):
+            return False
```
benauriemma: Think this should be True!

rackcity/utils/change_planner_utils.py
```diff
@@ -289,13 +298,69 @@ def get_page_count_response_for_cp(request, change_plan):
...
+            return True
+    return False
+
+
+def get_changes_on_asset(asset, asset_cp):
     fields = [field.name for field in Asset._meta.fields]
     changes = []
     for field in fields:
-        if getattr(asset, field) != getattr(assetCP, field):
+        if getattr(asset, field) != getattr(asset_cp, field):
```
benauriemma: Nice catch

rackcity/utils/change_planner_utils.py
juliasaveliff: oh


my






god

juliasaveliff: except hostnames & asset numbers could have changed in change plan so i dont love that for identifying an asset actually. I stand by my long and winding path

rackcity/utils/change_planner_utils.py
juliasaveliff: so True haah

rackcity/utils/change_planner_utils.py
benauriemma: One could make the argument that if one disconnected host1 from host2 and connected it to host3 but then renamed host2 to host3 and renamed host3 to host2 that they actually did nothing at all

benauriemma: But I stand with you.

rackcity/utils/change_planner_utils.py
benauriemma: Laughed audibly

benauriemma: Though I'm not proud of that

PR #372, opened 2020-03-30
Catch uncaught error on execute endpoint

juliasaveliff:


rackcity/views/change_plan_views.py
```diff
@@ -423,6 +423,17 @@ def change_plan_execute(request, id):
...
         if response:
             return response
+    else:
+        return JsonResponse(
+            {
+                "failure_message":
+                    Status.ERROR.value +
+                    GenericFailure.INTERNAL.value,
+                "errors":
+                    "Must specify change plan id as query param on execute"
```
juliasaveliff: this should never happen from the front end but there is a bug where it is happening so I am doing this so the error is at least caught :*

clairecfu: did you fix the front end bug? /what was the front end bug?

PR #373, opened 2020-03-30
cannot take admin away from admin user

juliasaveliff:


rackcity/views/user_views.py
```diff
@@ -391,16 +391,28 @@ def user_set_groups(request):
             },
             status=HTTPStatus.BAD_REQUEST,
         )
-
+    if (
+        (user.username == 'admin')
+        and ('admin' in data)
+        and (not (data['admin']))
+    ):
```
benauriemma: Should you be able to do anything to admin's permission?  I think it should just reject this regardless

juliasaveliff: MEH okay good point

PR #387, opened 2020-04-07
Add change plan button disabled fix

clairecfu:


src/utils/permissionUtils.tsx
```diff
@@ -0,0 +1,25 @@
+import { ElementType } from "./utils";
+
+export interface PermissionState {
+  model_management: boolean;
+  asset_management: boolean;
+  power_control: boolean;
+  audit_read: boolean;
+  admin: boolean;
+  datacenter_permissions: Array<number>;
+}
```
clairecfu, line 10: moved PermissionState to this new utils file

src/utils/permissionUtils.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    permissions.admin ||
+    (elementType === ElementType.DATACENTER && permissions.asset_management) ||
+    (elementType === ElementType.MODEL && permissions.model_management) ||
+    (elementType === ElementType.ASSET &&
+      (permissions.asset_management ||
+        permissions.datacenter_permissions.length > 0)) ||
+    elementType === ElementType.CHANGEPLANS
+  );
+}
```
clairecfu, line 25: new permission util function -> i use this in deciding whether or not to disable add button + bulk import button

src/components/elementView/elementTab.tsx
```diff
@@ -425,15 +435,9 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
...
-                        (this.props.element === ElementType.MODEL &&
-                          this.props.permissionState.model_management) ||
-                        (this.props.element === ElementType.ASSET &&
-                          this.props.permissionState.asset_management) ||
-                        (this.props.element === ElementType.ASSET &&
-                          this.props.permissionState.datacenter_permissions
-                            .length > 0)
+                    : !hasAddElementPermission(
+                        this.props.element,
+                        this.props.permissionState
```
clairecfu, line 440: refactor #1 bulk upload

src/components/elementView/elementTab.tsx
```diff
@@ -549,17 +553,9 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
...
-                      (this.props.element === ElementType.MODEL &&
-                        this.props.permissionState.model_management) ||
-                      (this.props.element === ElementType.ASSET &&
-                        this.props.permissionState.asset_management) ||
-                      (this.props.element === ElementType.ASSET &&
-                        this.props.permissionState.datacenter_permissions
-                          .length > 0)
+                  : !hasAddElementPermission(
+                      this.props.element,
+                      this.props.permissionState
```
clairecfu, line 558: refactor #2 add button

src/components/elementView/elementTab.tsx
```diff
@@ -425,15 +435,9 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
                 disabled={
                   this.props.changePlan
                     ? true
```
benauriemma, line 437: This means if you're in a change plan, you can't bulk upload anything - that's desired, right?

src/components/elementView/elementTab.tsx
```diff
@@ -549,17 +553,9 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
                 this.props.element !== ElementType.ASSET &&
                 this.props.changePlan
                   ? true
```
benauriemma, line 555: This means you can't add a user or datacenter if you're in a change plan, is that desired?  I think that's inconsistent with Racks isn't it?  In that you can add a rack but there's a warning that it makes those changes live

benauriemma: Also can you edit or delete datacenters, users, or models if you're in a change plan? I think it should be the same for all the write operations

src/components/elementView/elementTab.tsx
benauriemma: Is this function only for add/bulk upload, not edit/decommission/delete because the latter three require checking datacenter? I don't love that we factored this out but not that, but it isn't a huge deal

clairecfu: I only did add/bulk upload bc i only refactored elementTab, i can do a full refactor later

src/components/elementView/elementTab.tsx
clairecfu: yep

src/components/elementView/elementTab.tsx
clairecfu: nope, this is expected behavior for change plan, and was reflects the logic we had for ev3. the only reason why you're allowed to add a rack in a change plan is theoretically you might need to place a change plan asset on a rack that doesn't exist now. we don't allow this anywhere else, which is why we only have a warning for racks

src/components/elementView/elementTab.tsx
clairecfu: ok i lied i will do this refactor now

benauriemma:
Made tickets for all the questions I had here

PR #388, opened 2020-04-07
new custom asset fields +migration

clairecfu:
adding custom fields for custom asset on AbstractAsset


rackcity/models/asset.py
```diff
@@ -131,6 +133,13 @@ class AbstractAsset(AssetID):
     )
     comment = models.TextField(null=True, blank=True)
 
+    cpu = models.CharField(max_length=150, blank=True)
+    display_color = models.CharField(
+        max_length=7, validators=[validate_display_color], blank=True
+    )
+    storage = models.CharField(max_length=150, blank=True)
+    memory_gb = RCPositiveIntegerField(null=True, blank=True)
+
```
clairecfu, line 142: following up on our convo about how its bad practice to have null = True for CharFields, so i did not include that here. but important to track that this is different from how these fields are implemented in it_model

rackcity/models/it_model.py
```diff
@@ -1,26 +1,16 @@
...
+from rackcity.models.model_utils import DEFAULT_DISPLAY_COLOR,validate_display_color
 
 
 class ITModel(models.Model):
     vendor = models.CharField(max_length=150)
     model_number = models.CharField(max_length=150)
     height = models.PositiveIntegerField()
     display_color = models.CharField(
-        max_length=7, default="#394B59", validators=[validate_display_color],
+        max_length=7, default=DEFAULT_DISPLAY_COLOR, validators=[validate_display_color],
```
clairecfu, line 13: NO MAGIC VALUES!!!!!1

rackcity/models/it_model.py
```diff
@@ -1,26 +1,16 @@
...
-
-DEFAULT_DISPLAY_COLOR = "#394B59"
-
-
-def validate_display_color(value):
-    if value:
-        color_pattern = re.compile("#[A-Fa-f0-9]{6}")
-        if color_pattern.fullmatch(value) is None:
-            raise ValidationError(value + " is not a valid hex color")
+from rackcity.models.model_utils import DEFAULT_DISPLAY_COLOR,validate_display_color
```
benauriemma, line 5: `DEFAULT_DISPLAY_COLOR,validate_display_color` did you lint 🧐

rackcity/models/asset.py
benauriemma: Did you test what happens if you create an asset where these are null?  Does it default to empty string or does it throw an error?  I think the former is preferred but this seems like it would do the latter

benauriemma: Also could you change this in ITModel while you're at it?

rackcity/models/it_model.py
benauriemma: 😍

rackcity/models/it_model.py
clairecfu: BEN LINTER STRIKES AGAIN

rackcity/models/asset.py
clairecfu: If True, Django will store empty values as NULL in the database. Default is False.
Avoid using null on string-based fields such as CharField and TextField because empty string values will always be stored as empty strings, not as NULL. If a string-based field has null=True, that means it has two possible values for "no data": NULL, and the empty string. In most cases, it’s redundant to have two possible values for "no data"; the Django convention is to use the empty string, not NULL.

https://stackoverflow.com/questions/8609192/differentiate-null-true-blank-true-in-django

clairecfu: I believe it defaults to empty string and i did not get any warnings when i migrated but i will double check this

rackcity/models/asset.py
```diff
@@ -131,6 +133,13 @@ class AbstractAsset(AssetID):
     )
     comment = models.TextField(null=True, blank=True)
 
+    cpu = models.CharField(max_length=150, default="", blank=True)
+    display_color = models.CharField(
+        max_length=7, validators=[validate_display_color], default="", blank=True
+    )
+    storage = models.CharField(max_length=150, blank=True, default="")
+    memory_gb = RCPositiveIntegerField(null=True, blank=True)
+
```
clairecfu, line 142: adding default values

PR #390, opened 2020-04-08
Modify bug fix 

clairecfu:
- [x] modifying asset was throwing non unique hostname error when you were modifying any field on asset

... suspected that this affected all modifications of models


rackcity/utils/asset_utils.py
```diff
@@ -410,6 +410,7 @@ def save_all_connection_data(data, asset, user, change_plan=None):
 
 
 def save_all_field_data(data, asset, change_plan=None):
+    id = data[id]
```
clairecfu, line 413: id was never defined, i guess it defaulted to some built in django value

rackcity/utils/asset_utils.py
```diff
@@ -410,6 +410,7 @@ def save_all_connection_data(data, asset, user, change_plan=None):
 
 
 def save_all_field_data(data, asset, change_plan=None):
+    id = data["id"]
```
juliasaveliff, line 413: nice catch

PR #391, opened 2020-04-08
Database changes for blades, plus some temporary fixes to endpoints

benauriemma:


rackcity/api/serializers/it_model_serializers.py
```diff
@@ -5,7 +5,9 @@
 
 
 class ITModelSerializer(serializers.ModelSerializer):
-
+    height = RCIntegerField(
+        allow_null=True, max_value=2147483647, min_value=0, required=False
+    )
```
clairecfu, line 10: Can you explain why you added this to the serializer?

rackcity/models/asset.py
```diff
@@ -121,13 +121,34 @@ def validate_owner(value):
...
+
+
 class AssetID(models.Model):
     id = models.AutoField(primary_key=True)
 
 
 class AbstractAsset(AssetID):
-    rack_position = models.PositiveIntegerField()
-
+    rack_position = RCPositiveIntegerField(null=True, blank=True,)
```
clairecfu, line 150: whats the diff between RCPositiveIntegerField and PostiveIntegerField?

rackcity/models/asset.py
```diff
@@ -161,7 +182,18 @@ class Asset(AbstractAsset):
...
-        Rack, on_delete=models.CASCADE, verbose_name="related rack",
+        Rack,
+        on_delete=models.CASCADE,
+        verbose_name="related rack",
+        null=True,
+        blank=True,
+    )
+    chassis = models.ForeignKey(
+        "self",
+        on_delete=models.PROTECT,
```
clairecfu, line 193: shouldn't the on_delete behavior for chassis and rack be consistent?

rackcity/models/asset.py
```diff
@@ -250,7 +289,18 @@ class AssetCP(AbstractAsset):
...
+        verbose_name="blade chassis",
+        null=True,
+        blank=True,
```
clairecfu, line 303: I think `on_delete` should be SET_NULL?

rackcity/views/asset_views.py
```diff
@@ -193,25 +193,33 @@ def asset_add(request):
         validate_user_asset_permission_to_add(request.user, serializer.validated_data)
     except UserAssetPermissionException as auth_error:
         return JsonResponse(
-            {"failure_message": Status.AUTH_ERROR + str(auth_error)},
+            {"failure_message": Status.AUTH_ERROR.value + str(auth_error)},
             status=HTTPStatus.UNAUTHORIZED,
         )
     except Exception as error:
+        print("this error")
```
clairecfu, line 200: can u remove this print statement  - CLAIRE LINTER

rackcity/api/serializers/it_model_serializers.py
benauriemma: Yep - we use RCIntegerField (rather than the default serializers.IntegerField) for all optional numeric fields to handle empty strings being passed from forms

rackcity/models/asset.py
benauriemma: Same as above - handles empty strings better, which is only necessary now that it's optional

rackcity/models/asset.py
benauriemma: You shouldn't be able to delete either a rack or a chassis that has assets on it.  We check this manually on rack (and might end up checking it manually on chassis, likely for a better error message - I have a task to track this) but to be safe since I haven't implemented that check yet, I chose `PROTECT` so that the database won't allow a chassis with blades to be deleted.  We didn't know what `on_delete` did when we wrote it for Racks

rackcity/models/asset.py
benauriemma: This reference is to another AssetCP, so I think we can enforce within the change plan that the chassis (in the CP) won't be deleted if it has blades (in the CP) in it

rackcity/views/asset_views.py
benauriemma: Damn it
Claire linter - 1
Ben linter - 1

rackcity/models/asset.py
clairecfu: oof ok more of a side note then: when detecting conflicts on change plan, every blade is going to need to check if it's chassis has any conflicts (specifically if the live chassis has been deleted)

PR #394, opened 2020-04-08
Database migrations for Site model

juliasaveliff:
Changes in this PR 
- Rename Datacenter to Site and add `is_storage` field
- Rename `datacenter_permissions` to `site_permissions`
- Update any code that would break 
- Add `offline_storage_site` to AbstractAsset and update `validate_location_type` method


rackcity/api/objects/rack_range.py
```diff
@@ -1,11 +1,11 @@
 from rest_framework import serializers
 from rackcity.models.rack import validate_row_letter
-from rackcity.models import Datacenter
+from rackcity.models import Site
 
 
 class RackRangeSerializer(serializers.Serializer):
     datacenter = serializers.PrimaryKeyRelatedField(
-        label="datacenter", queryset=Datacenter.objects.all(), required=False
+        label="datacenter", queryset=Site.objects.all(), required=False
```
clairecfu, line 8: are you waiting to make changes to rack range serializer?

rackcity/models/asset.py
```diff
@@ -121,24 +122,34 @@ def validate_owner(value):
...
+                "Blade assets in storage must not have a chassis or chassis slot. "
+            )
+        if (not offline_storage_site) and (not chassis or not chassis_slot):
+            raise ValidationError(
+                "Blade assets not in storage must have a chassis and chassis slot. "
+            )
         if rack or rack_position:
             raise ValidationError(
-                "Rackmount assets must not have a rack or rack position. "
+                "Blade assets must not have a rack or rack position. "
```
clairecfu, line 152: good catch

rackcity/api/objects/rack_range.py
juliasaveliff: TBD on that field, might actually keep it datacenter because a rack will always have a datacenter never an offline storage site

PR #403, opened 2020-04-09
custom asset changes

clairecfu:
salvaged all my changes 

backend
- [x] adding custom fields to serializer 
- [x] adding serializer validation for memory_gb 
- [x] made some migrations but then made new migrations that undid old migrations 
front end
- [x] Removing legacy code from `utils.tsx`
- [x] adding custom model fields to asset form
- [x] updating asset detail view to show changed properties

### DIFF
--- a/src/components/elementView/detailedView/propertiesView.tsx
+++ b/src/components/elementView/detailedView/propertiesView.tsx
@@ -10,7 +10,6 @@ import {
   isModelObject,
   isObject,
   ModelFieldsTable,
-  ROUTES,
   ChangePlan,
 } from "../../../utils/utils";
 import "./propertiesView.scss";
@@ -20,12 +19,12 @@ export interface AlertState {
   isDeleteOpen: boolean;
   fields: Array<string>;
 }
-// var console: any = {};
-// console.log = function() {};
 
 interface PropertiesViewProps {
   data: ElementObjectType;
   changePlan: ChangePlan;
+  title?: string;
+  data_override?: any;
 }
 
 class PropertiesView extends React.PureComponent<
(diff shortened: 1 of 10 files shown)


src/components/elementView/detailedView/propertiesView.tsx
```diff
@@ -75,20 +80,7 @@ class PropertiesView extends React.PureComponent<
           dat = <p>{network_ports.toString()}</p>;
         }
       } else if (item === "model") {
-        const isDecommissioned = data["decommissioning_user"];
```
brendancheng20, line 78: What's the reason for getting rid of decommissioning stuff here?

clairecfu: i removed the onClick logic from the model property because I added a button in assetView.tsx that links to model view page, so i dont need this logic anymore

clairecfu: (the decommissioned logic is preserved for the button)

src/utils/utils.tsx
```diff
@@ -63,7 +56,7 @@ export interface AssetCPObject extends AssetObject {
 }
 interface ParentAssetObject extends ElementObject {
   asset_number: string;
-  hostname: string;
+  hostname: string|null;
```
brendancheng20, line 59: Does this require a white space? Like string | null?

src/utils/utils.tsx
```diff
@@ -72,6 +65,10 @@ interface ParentAssetObject extends ElementObject {
   comment: string;
   decommissioning_user?: string;
   time_decommissioned?: string;
+  cpu:string;
+  storage:string;
+  display_color:string;
+  memory_gb:string |null;
```
brendancheng20, line 71: Similar question about whitespace. Not a big deal tho, just curious if we need that

PR #406, opened 2020-04-10
Blades: Model front end changes 

clairecfu:
- [x] added "Mount Type" to table and detail view 
- [x] added "Mount Type" to model form -> model form should be done for blades 
- [x] fixed bug where continued requests are made if no vendors exist 
- [x] fixed bug where model detail view crashes if no existing assets


src/utils/utils.tsx
```diff
@@ -197,10 +200,17 @@ export interface DatacenterObject extends ElementObject {
   abbreviation: string;
 }
 
+export enum MountTypes {
+  RACKMOUNT = "Asset",
+  BLADE_CHASSIS = "Blade Chassis",
+  BLADE = "Blade Server ",
+}
+
```
clairecfu, line 208: this mirrors the strings on the backend. Even though i just changed these i am not super happy with these... after reading the requirements i think it should be "Normal Rackmount, Blade Chassis, Blade", which mirrors the vocabulary in the requirements

src/utils/utils.tsx
```diff
@@ -115,6 +115,7 @@ export const ModelFieldsTable: any = {
   memory_gb: "Memory (GB)",
   storage: "Storage",
   comment: "Comment",
+  model_type: "Mount Type"
```
clairecfu, line 118: similarly, i think this should be called Mount Type, not model type, but keeping variable names consistent with backend for now

src/forms/modelForm.tsx
```diff
@@ -189,163 +178,194 @@ class ModelForm extends React.Component<ModelFormProps, ModelFormState> {
...
+                value={values.display_color ? values.display_color : "#394B59"}
+                onChange={this.handleChange}
+              />
+            </FormGroup>
+          </Card>
+          <Card>
+            <RadioGroup
+              label="Mount Type*"
+              className="radio-group"
+              onChange={(e: any) => {
```
benauriemma, line 248: This is picky but I'm not sure Mount Type is accurate bc both classic assets and blade chassis are rackmount.  Maybe model type?  Or is there another word that describes it better

benauriemma: Discussed in standup that'll we'll leave it but ask Bletsch to confirm - are you gonna post on Piazza?

src/forms/modelForm.tsx
```diff
@@ -189,163 +178,194 @@ class ModelForm extends React.Component<ModelFormProps, ModelFormState> {
...
+            >
+              <Radio
+                label={MountTypes.RACKMOUNT}
+                value={MountTypes.RACKMOUNT}
+              />
+              <Radio
+                label={MountTypes.BLADE_CHASSIS}
+                value={MountTypes.BLADE_CHASSIS}
+              />
+              <Radio label={MountTypes.BLADE} value={MountTypes.BLADE} />
```
benauriemma, line 265: My new least favorite thing the linter has done

src/utils/utils.tsx
benauriemma: You're tracking changing this assuming Bletsch confirms?

clairecfu: yeet

src/forms/modelForm.tsx
clairecfu: hehehehehehehe

src/forms/modelForm.tsx
clairecfu: yep

src/utils/utils.tsx
benauriemma: I assume you're tracking

PR #407, opened 2020-04-10
Sites (offline + datacenter) endpoint update

juliasaveliff:
Changes in this PR: 
- [x] Rename datacenter serializer and add `is_storage` field
- [x] Rename datacenter views
- [x] Create offline storage site views
- [x] Update endpoint URLS in the front end

Related changes to do in future PR:
- [ ] Update/create related assets endpoints
- [ ] Put offline sites on front end


rackcity/views/site_views.py
```diff
@@ -51,175 +64,200 @@ def datacenter_create(request):
...
-        },
+        {"success_message": Status.SUCCESS.value + new_site.abbreviation + " created"},
         status=HTTPStatus.CREATED,
     )
 
 
 @api_view(["POST"])
 @permission_required(PermissionPath.ASSET_WRITE.value, raise_exception=True)
-def datacenter_delete(request):
+def site_modify(request):
```
juliasaveliff, line 87: modify endpoint was moved from bottom of file to here so it looks like it replaced the delete endpoint but that's a LIE - might be easier to look at changes in PyCharm

rackcity/views/site_views.py
```diff
@@ -51,175 +64,205 @@ def datacenter_create(request):
...
+            site_with_name = None
+            if field == "name":
+                site_with_name = Site.objects.filter(name__iexact=data[field])
+            elif field == "abbreviation":
+                site_with_name = Site.objects.filter(abbreviation__iexact=data[field])
+            if (
+                site_with_name
+                and (len(site_with_name) > 0)
+                and (site_with_name[0].id != id)
+            ):
```
juliasaveliff, line 135: changed logic slightly because `is_storage` is also a field

rackcity/views/site_views.py
```diff
@@ -51,175 +64,205 @@ def datacenter_create(request):
...
-                "errors": "No existing datacenter with id=" + str(id),
+                "errors": "No existing site with id=" + str(id),
             },
             status=HTTPStatus.BAD_REQUEST,
         )
-    dc_abbreviation = existing_dc.abbreviation
-
+    for field in data.keys():
+        if field != "id":
+            if (isinstance(data[field], str)) and (len(data[field]) == 0):
```
juliasaveliff, line 116: changed logic slightly because `is_storage` is also a field

src/components/elementView/detailedView/assetView/assetView.tsx
```diff
@@ -180,7 +180,7 @@ export class AssetView extends React.PureComponent<
     const headers = getHeaders(this.props.token);
 
     axios
-      .post(API_ROOT + "api/datacenters/get-many", {}, headers)
+      .post(API_ROOT + "api/sites/datacenters/get-many", {}, headers)
```
juliasaveliff, line 183: only change in this file - rest is linting

src/components/elementView/elementTab.tsx
```diff
@@ -304,16 +307,14 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
...
-      .post(API_ROOT + "api/datacenters/add", dc, headers)
-      .then((res) => {
-        this.handleDataUpdate(true);
-        this.handleClose();
-        this.addSuccessToast(res.data.success_message);
-        if (this.props.updateDatacenters) {
-          this.props.updateDatacenters();
-        }
-      });
+    return axios.post(API_ROOT + "api/sites/add", dc, headers).then((res) => {
```
juliasaveliff, line 310: this is linting

rackcity/models/site.py
```diff
@@ -5,3 +5,11 @@ class Site(models.Model):
     abbreviation = models.CharField(max_length=6, unique=True)
     name = models.CharField(max_length=150, unique=True)
     is_storage = models.BooleanField(default=False)
+
+    @staticmethod
```
benauriemma, line 9: What's this decorator for?

rackcity/views/site_views.py
```diff
@@ -51,175 +64,205 @@ def datacenter_create(request):
...
-                + GenericFailure.ON_SAVE.value,
+                "failure_message": Status.DELETE_ERROR.value
+                + "Site"
+                + GenericFailure.ON_DELETE.value,
                 "errors": str(error),
             },
             status=HTTPStatus.BAD_REQUEST,
         )
-    log_action(request.user, existing_dc, Action.MODIFY)
+    log_delete(request.user, ElementType.SITE, existing_site.abbreviation)
```
benauriemma, line 228: `ElementType.SITE` -> `ElementType.SITE.value`!

benauriemma: Wanna check that this doesn't happen elsewhere

benauriemma: I was wrong but you're crazy

rackcity/views/site_views.py
benauriemma: Github is brazy

rackcity/models/site.py
juliasaveliff: if it's a static method, you can access it from the class without needing a reference to an object of the class to call it on (so the same way we're doing `Site.objects.all()` anywhere we want we can also do `Site.get_datacenters()`)

rackcity/views/site_views.py
juliasaveliff: https://www.youtube.com/watch?v=k85mRPqvMbE

PR #410, opened 2020-04-11
location validation to asset add and modify for blades

benauriemma:


rackcity/views/asset_views.py
```diff
@@ -202,23 +203,31 @@ def asset_add(request):
             status=HTTPStatus.BAD_REQUEST,
         )
 
-    if (
-        serializer.validated_data["model"].is_rackmount()
-        and "rack" in serializer.validated_data
-        and "rack_position" in serializer.validated_data
```
benauriemma, line 208: The form should never let this happen

rackcity/views/asset_views.py
```diff
@@ -635,7 +644,8 @@ def asset_bulk_upload(request):
                 )
             model = ITModel.objects.get(id=asset_data["model"])
             try:
-                validate_asset_location(
+                # TODO: add chassis validation to bulk
```
benauriemma, line 647: I have a ticket for this

benauriemma:
https://www.notion.so/672c624b578946508f13a4c2da4cb324?v=0bd36cfb32604a0b8a9313569f70194b

benauriemma:
I'm admittedly not confident in the change plan work here, and am not sure that my use cases for change plan are thorough. Would still love a crash course in change plan stuff because I don't really feel confident making changes to it at the moment

rackcity/utils/change_planner_utils.py
```diff
@@ -75,22 +78,44 @@ def detect_conflicts_cp(sender, **kwargs):
...
+                )
+            except LocationException:
+                assetcp.asset_conflict_location = asset
+                AssetCP.objects.filter(id=assetcp.id).update(
+                    asset_conflict_location=asset
+                )
+    else:
+        asset.location_conflict.clear()
+        for assetcp in AssetCP.objects.filter(
+            Q(chassed=asset.chassis_id)
```
clairecfu, line 105: should this be chassis not chassed

rackcity/utils/change_planner_utils.py
```diff
@@ -75,22 +78,44 @@ def detect_conflicts_cp(sender, **kwargs):
...
-                assetcp.rack_position,
-                assetcp.model.height,
-                asset_id=assetcp.id,
-            )
-        except LocationException:
-            assetcp.asset_conflict_location = asset
-            AssetCP.objects.filter(id=assetcp.id).update(asset_conflict_location=asset)
+
+    if asset.model.is_rackmount():
+        asset.location_conflict.clear()
```
clairecfu, line 83: small nit but move this line outside of if statement bc we need it in both cases

rackcity/views/asset_views.py
clairecfu: so data will always have a field "rack" but it can sometimes be null if the user does not select a rack -> hence leading to us not catching that error. im tracking that bug fix though

clairecfu: wait actually, if you want to take over that bug? because I imagine we would want to add similar logic for if a user forgets to select a chassis for a blade slot

clairecfu: doesnt need to be done in this PR just flagging it here

rackcity/views/asset_views.py
```diff
@@ -202,23 +203,31 @@ def asset_add(request):
...
                 rack_id, rack_position, height, change_plan=change_plan
             )
         except LocationException as error:
             return JsonResponse(
                 {"failure_message": Status.CREATE_ERROR.value + str(error)},
                 status=HTTPStatus.BAD_REQUEST,
             )
+    else:
+        chassis_id = serializer.validated_data["chassis"].id
+        chassis_slot = serializer.validated_data["chassis_slot"]
```
clairecfu, line 221: this could potentially be null. same deal as rack

rackcity/utils/change_planner_utils.py
benauriemma: Oof

rackcity/utils/change_planner_utils.py
benauriemma: Will do

PR #411, opened 2020-04-11
Fixing bugs from refactor (mainly change plan)

clairecfu:


rackcity/models/asset.py
```diff
@@ -100,18 +100,21 @@ def validate_asset_number_uniqueness(value, asset_id, change_plan, related_asset
...
     if (len(matching_assets) > 0) and (
         not (related_asset and matching_assets[0].id == related_asset_id)
     ):
+        print(matching_assets[0].id, related_asset_id)
         raise ValidationError(
             "'"
             + str(value)
             + "'is not a unique asset number. \
-            An existing asset on exists with this asset number."
+            An existing asset exists with this asset number."
```
clairecfu, line 117: typo

rackcity/permissions/permissions.py
```diff
@@ -85,6 +85,9 @@ def validate_user_asset_permission_to_add(user, validated_data):
     # TODO: make this check work for blades, which don't have a rack
     if "rack" not in validated_data:
         return
+    ## ading this to throw readable error if no rack specified
+    if not validated_data["rack"]:
+        raise Exception("Rack must be specified")
```
clairecfu, line 90: adding a readable error message for when no rack is selected on asset creation.

rackcity/utils/asset_utils.py
```diff
@@ -54,18 +54,18 @@ def get_existing_power_port(port_name, asset_id, change_plan=None):
         return power_port
 
 
-def get_or_create_asset_cp(asset_id, change_plan):
+def does_asset_exist(asset_id, change_plan):
     try:
-        existing_asset_cp = AssetCP.objects.get(
-            asset=asset_id, change_plan=change_plan.id
+        AssetCP.objects.get(
+            id=asset_id, change_plan=change_plan.id
```
clairecfu, line 60: does not have a field asset

rackcity/utils/asset_utils.py
```diff
@@ -54,18 +54,18 @@ def get_existing_power_port(port_name, asset_id, change_plan=None):
         return power_port
 
 
-def get_or_create_asset_cp(asset_id, change_plan):
+def does_asset_exist(asset_id, change_plan):
```
clairecfu, line 57: since this function is only used once, i modified it to better suit its purpose

rackcity/utils/asset_utils.py
```diff
@@ -81,7 +81,7 @@ def get_or_create_asset_with_hostname(hostname, change_plan=None):
             assets, assets_cp = get_assets_for_cp(change_plan.id)
             if assets.filter(hostname=hostname).exists():
                 asset_live = assets.get(hostname=hostname)
-                asset = copy_asset_to_new_asset_cp(asset_live, hostname, change_plan)
+                asset = copy_asset_to_new_asset_cp(asset_live,change_plan)
```
clairecfu, line 84: just a bug, copy_asset_to_new_asset_cp only takes 2 params

rackcity/utils/asset_utils.py
```diff
@@ -409,8 +408,9 @@ def save_all_connection_data(data, asset, user, change_plan=None):
     return warning_message
 
 
-def save_all_field_data(data, asset, change_plan=None):
+def save_all_field_data(data, asset, change_plan=None, create_asset_cp = False):
```
clairecfu, line 411: create_asset_cp will be True if you are on a change plan AND are modifying an aset on this change plan for the first time

rackcity/utils/asset_utils.py
```diff
@@ -456,16 +461,28 @@ def save_all_field_data(data, asset, change_plan=None):
...
-        setattr(asset, field, value)
-        try:
+
+        print(field, value)
+        if field is not "id":
+            setattr(asset, field, value)
+
+    try:
+        if create_asset_cp:
+            asset_cp = copy_asset_to_new_asset_cp(asset, change_plan)
```
clairecfu, line 479: make all the modifications on the original live asset, but dont save it-> copy values to a new assetcp

rackcity/utils/rackcity_utils.py
```diff
@@ -73,10 +73,12 @@ def validate_asset_location(
...
     for asset_in_rack in assets.filter(rack=rack_id):
         # Ignore if asset being modified conflicts with its old location
+
         is_valid_conflict = asset_id is None or asset_in_rack.id != asset_id
-        if change_plan:
+        print("location", is_valid_conflict,asset_in_rack.id, asset_id)
+        if related_asset_id:
             is_valid_conflict = (
-                related_asset_id is not None and asset_in_rack.id != related_asset_id
+                asset_in_rack.id != related_asset_id
```
clairecfu, line 81: if you are on a change plan but there is not related_asset_id, then we are still validating the location of the change on the "live asset"

rackcity/views/asset_views.py
```diff
@@ -335,13 +350,15 @@ def asset_modify(request):
...
-    failure_message = save_all_field_data(data, existing_asset, change_plan=change_plan)
+    print("Creating new asset cp",create_new_asset_cp)
+    (asset_cp,failure_message) = save_all_field_data(data, existing_asset, change_plan=change_plan,create_asset_cp = create_new_asset_cp)
     if failure_message:
         return JsonResponse(
             {"failure_message": Status.MODIFY_ERROR.value + failure_message},
             status=HTTPStatus.BAD_REQUEST,
         )
+    if asset_cp:
+        existing_asset = asset_cp
```
clairecfu, line 361: need to get the new assetCP now because  the rest of the logic assumes that it is getting an assetcp

rackcity/utils/asset_utils.py
juliasaveliff: might make more sense here to use `.exists()` on the query? since we don't actually want the object being returned? so 
```
return (AssetCP.objects.filter(id=asset_id, change_plan=change_plan.id).exists()) or (Asset.objects.filter(id=asset_id).exists())
```

rackcity/permissions/permissions.py
juliasaveliff: this doesn't really have to do with permissions, since it just means the request being sent is invalid, so I think it should go in asset_add method and not here

rackcity/utils/asset_utils.py
```diff
@@ -54,18 +54,18 @@ def get_existing_power_port(port_name, asset_id, change_plan=None):
...
     except ObjectDoesNotExist:
         try:
             existing_asset_live = Asset.objects.get(id=asset_id)
         except ObjectDoesNotExist:
-            return None
-        existing_asset_cp = copy_asset_to_new_asset_cp(existing_asset_live, change_plan)
-    return existing_asset_cp
+            return False
+        # existing_asset_cp = copy_asset_to_new_asset_cp(existing_asset_live, change_plan)
+    return True;
```
juliasaveliff, line 68: semi colon tehe

rackcity/utils/asset_utils.py
juliasaveliff: nice catch

rackcity/views/asset_views.py
```diff
@@ -283,14 +289,22 @@ def asset_modify(request):
...
+
+        create_new_asset_cp = not AssetCP.objects.filter(
+            id=id, change_plan=change_plan.id
+        ).exists()
+
+        if not create_new_asset_cp:
+            existing_asset =  AssetCP.objects.get(
+            id=id, change_plan=change_plan.id
+        )
+        if not does_asset_exist(id, change_plan):
```
juliasaveliff, line 307: I like the new method, but a little confused about this logic? worth talking about later

rackcity/utils/asset_utils.py
```diff
@@ -456,16 +461,28 @@ def save_all_field_data(data, asset, change_plan=None):
...
+            asset_cp.save()
+            return asset_cp, None
+
+        else:
             asset.save()
-        except Exception as error:
-            return parse_save_validation_error(error, "Asset")
+            return None, None
+    except Exception as error:
+        return parse_save_validation_error(error, "Asset")
```
juliasaveliff, line 488: it looks like you should be returning a tuple everywhere - but a general comment on that - I think it would be more ideal to separate these different paths of logic into separate methods because one of my goals of the first refactor was to make each path more obvious and avoid a lot of if this, do all this, but if not, do all this, etc - if that makes sense? we can talk about this later too

rackcity/views/asset_views.py
juliasaveliff: see comment above about that method change

rackcity/permissions/permissions.py
clairecfu: removing this and sending this to ben so he can also handle similar logic if no chassis is specified

rackcity/utils/asset_utils.py
clairecfu: done

rackcity/utils/asset_utils.py
clairecfu: cool, i have thoughts on how to fix this but want to run by you before i do it

rackcity/utils/asset_utils.py
```diff
@@ -409,63 +400,94 @@ def save_all_connection_data(data, asset, user, change_plan=None):
     return warning_message
 
 
-def save_all_field_data(data, asset, change_plan=None):
-    id = data["id"]
+def save_all_field_data_live(data, asset):
+    asset_id = data["id"]
```
clairecfu, line 404: renaming id variables to asset_id bc id is a built in variable, bad practice

clairecfu:
> it looks good to me, except there are some warnings in PyCharm about existing_asset & asset_cp being referenced before assignment - are you seeing those?

instantiated existing_asset and asset_cp -> i think existing_asset would never actually not be assigned, but asset_cp is a good catch

PR #414, opened 2020-04-11
Paging bug fix

juliasaveliff:
https://www.notion.so/fa565cfaee7a460bbf801045cca585b8?v=b318357f12804f9194b85df75248608b&p=fe905c0b0dda43aeb7199651d5aeb805

### DIFF
--- a/src/components/elementView/elementTab.tsx
+++ b/src/components/elementView/elementTab.tsx
@@ -183,7 +183,7 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
       ? API_ROOT + "api/sites/datacenters/pages"
       : API_ROOT + "api/" + path + "/pages";
     return axios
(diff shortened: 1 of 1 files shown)


src/components/elementView/elementTab.tsx
```diff
@@ -183,7 +183,7 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
       ? API_ROOT + "api/sites/datacenters/pages"
       : API_ROOT + "api/" + path + "/pages";
     return axios
-      .post(API_ROOT + url, { filters: filtersCopy }, config)
+      .post(url, { filters: filtersCopy }, config)
```
clairecfu, line 186: technicallllyyy would be cleaner + more consistent with how we do it else where to leave API_ROOT + url, but remove API_ROOT from the the variable url itself since that is always going to be there

PR #416, opened 2020-04-12
Chassis power API endpoints

juliasaveliff:
Done:
- [x] Update existing pdu power endpoints
- [x] Implement chassis power endpoints
Note: to test locally, each developer will need to add bcman credentials to local .env file (these are already added to heroku, so should work on dev) 

To do in another PR:
- [ ] Connection/timeout handling and testing


src/components/elementView/powerView/powerView.tsx
```diff
@@ -84,7 +84,7 @@ export class PowerView extends React.PureComponent<
       });
       axios
         .get(
-          API_ROOT + "api/power/get-state/" + this.props.asset!.id,
+          API_ROOT + "api/rack-power/status/" + this.props.asset!.id,
```
clairecfu, line 87: we need to check that an asset is rack powered before calling this endpoint and more generally add the new power logic to the front end  (ex. if asset = blade, don't show rack power details, if asset= chassis, show chassis power details)

PR #418, opened 2020-04-12
Update asset endpoints to return correct assets

juliasaveliff:
- Get many for racked assets (update endpoint)
- Get many for assets in offline storage (new endpoint)
- Page count for racked assets (update endpoint)
- Page count for assets in offline storage (new endpoint)
- Involves refactor to how we get all assets on live & on CP


rackcity/views/asset_views.py
```diff
@@ -784,7 +817,7 @@ def asset_bulk_approve(request):
         return JsonResponse({"warning_message": warning_message}, status=HTTPStatus.OK)
     else:
         return JsonResponse(
-            {"success_message": "Assets succesfully modified. "}, status=HTTPStatus.OK
+            {"success_message": "Assets successfully modified. "}, status=HTTPStatus.OK
```
benauriemma, line 820: 😳

rackcity/views/decommission_views.py
```diff
@@ -267,12 +269,14 @@ def decommissioned_asset_page_count(request):
     Return total number of pages according to page size, which must be
     specified as query parameter.
     """
-    (change_plan, failure_response) = get_change_plan(query_params.get("change_plan"))
+    (change_plan, failure_response) = get_change_plan(
+        request.query_params.get("change_plan")
+    )
```
benauriemma, line 274: Was this just a bug that we weren't calling `request.query_params`? Did we know about that?

juliasaveliff: yeah this was an uncaught bug. not sure when it got introduced. we could look at the git blame for fun

PR #419, opened 2020-04-12
Update table + properties detail view for blade/chassis assets 

clairecfu:
- [x] adding chassis + chassis slot to table and asset detail view
- [x] don't show propertiesview if asset is a blade 
- [x] updating serializers for blades (adding datacenter field, and making ChassisSerializer)


rackcity/api/serializers/asset_serializers.py
```diff
@@ -90,6 +95,24 @@ class Meta:
             "memory_gb",
         )
 
+class ChassisSerializer(serializers.ModelSerializer):
+    """
+    Serializers the information we want for a chassis that a blade is in (only used for serializing info to be sent)
+    """
+    rack = RackSerializer()
+
```
clairecfu, line 103: adding this so that I can get the detailed rack information I need for a chassis, but avoiding having an endless loop of serializers

src/components/elementView/elementTable.tsx
```diff
@@ -1143,15 +1136,24 @@ class ElementTable extends React.Component<
                                 event.stopPropagation();
                               }}
                             >
-                              {this.props.isDecommissioned  ? null : (
+                              {this.props.isDecommissioned ? null : (
```
benauriemma, line 1139: Had these files not been listed before?

PR #420, opened 2020-04-12
BUG FIX - creating users

juliasaveliff:
addresses two ev3 bugs:
image

### DIFF
--- a/rackcity/views/user_views.py
+++ b/rackcity/views/user_views.py
@@ -11,7 +11,7 @@ from rackcity.permissions.groups import (
     update_user_groups,
     update_user_site_permissions,
 )
-from rackcity.permissions.permissions import PermissionPath
+from rackcity.permissions.permissions import PermissionPath, RegisterUserPermission
 from rackcity.utils.query_utils import (
     get_page_count_response,
     get_many_response,
@@ -37,10 +37,9 @@ from rest_framework.parsers import JSONParser
 from rest_framework.permissions import IsAuthenticated
 
 
-class RegisterNameView(PermissionRequiredMixin, RegisterView):
+class RegisterNameView(RegisterView):
+    permission_classes = IsAuthenticated, RegisterUserPermission
     serializer_class = RegisterNameSerializer
-    permission_required = PermissionPath.USER_WRITE.value
-    raise_exception = True
 
 
 @api_view(["POST"])
--- a/src/components/elementView/elementTab.tsx
+++ b/src/components/elementView/elementTab.tsx
@@ -318,14 +318,22 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
   };
 
   private createUser = (user: CreateUserObject, headers: any): Promise<any> => {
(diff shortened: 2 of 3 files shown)


rackcity/views/user_views.py
```diff
@@ -37,10 +37,9 @@
 from rest_framework.permissions import IsAuthenticated
 
 
-class RegisterNameView(PermissionRequiredMixin, RegisterView):
+class RegisterNameView(RegisterView):
+    permission_classes = IsAuthenticated, RegisterUserPermission
     serializer_class = RegisterNameSerializer
-    permission_required = PermissionPath.USER_WRITE.value
-    raise_exception = True
```
juliasaveliff, line 43: the PermissionRequiredMixin was not working, so no requests to this endpoint were passing the permissions check. I have no idea why it was working for creating one user but not two in a row before our eval

rackcity/views/user_views.py
```diff
@@ -37,10 +37,9 @@
 from rest_framework.permissions import IsAuthenticated
 
 
-class RegisterNameView(PermissionRequiredMixin, RegisterView):
+class RegisterNameView(RegisterView):
+    permission_classes = IsAuthenticated, RegisterUserPermission
```
juliasaveliff, line 41: overwriting the permission_classes instead allows us to override some of the behavior happening in the `django-rest-auth` library

src/components/elementView/elementTab.tsx
```diff
@@ -318,11 +318,22 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
...
+        this.handleDataUpdate(true);
+        this.handleClose();
+        this.addSuccessToast("SUCCESS: User " + username + " created.");
+      })
+      .catch((err) => {
+        for (let key in err.response.data) {
+          let errors = err.response.data[key];
+          errors.forEach((message: string) => {
+            this.addErrorToast(message);
+          });
```
juliasaveliff, line 334: the errors weren't being caught as a toast before which explains why the 403 HTML page was showing up one character at a time. since the errors at this endpoint are thrown by `django-rest-auth` and not our code, we can't show `failure_message` as usual, so i'm parsing the list of errors that gets returned instead

src/components/elementView/elementTab.tsx
```diff
@@ -318,11 +318,22 @@ class ElementTab extends React.Component<ElementTabProps, ElementViewState> {
...
-      this.handleClose();
-      this.addSuccessToast(res.data.success_message);
-    });
+    const username = user.username;
+    return axios
+      .post(API_ROOT + "api/users/add", user, headers)
+      .then((res) => {
+        this.handleDataUpdate(true);
+        this.handleClose();
+        this.addSuccessToast("SUCCESS: User " + username + " created.");
```
juliasaveliff, line 327: again, responses from this endpoint are thrown by `django-rest-auth` and not our code, so `success_message` is never sent, so writing the success message here instead

PR #421, opened 2020-04-12
add validation on model save

benauriemma:


rackcity/models/asset.py
```diff
@@ -217,6 +217,7 @@ class Meta:
 
     def save(self, *args, **kwargs):
         try:
+            self.full_clean()
```
juliasaveliff, line 220: what is this?

rackcity/utils/change_planner_utils.py
```diff
@@ -76,21 +79,41 @@ def detect_conflicts_cp(sender, **kwargs):
...
-                asset.rack_id,
-                assetcp.rack_position,
-                assetcp.model.height,
-                asset_id=assetcp.id,
-            )
-        except LocationException:
-            assetcp.asset_conflict_location = asset
-            AssetCP.objects.filter(id=assetcp.id).update(asset_conflict_location=asset)
+    if asset.model.is_rackmount():
+        for assetcp in AssetCP.objects.filter(
```
juliasaveliff, line 83: would you mind renaming `assetcp` to `asset_cp`? it's more consistent with the rest of the code

rackcity/utils/errors_utils.py
```diff
@@ -161,6 +161,11 @@ def parse_save_validation_error(error, object_name):
     failure_detail = ""
     if isinstance(error, ValidationError):
         for err in error:
+            print(err)
```
juliasaveliff, line 164: print statement :P

rackcity/utils/change_planner_utils.py
benauriemma: I didn't write this but will do

rackcity/utils/errors_utils.py
benauriemma: Damnnnn it

rackcity/models/asset.py
benauriemma: `full_clean()` is essentially the validation method on a model -- it cleans the data for the database and will throw a ValidationError if any validation step fails.  We thought that it was called automatically on both serializer.save() and model.save(), but turns out it's only called on the serializer.  Read more about what it does here: https://docs.djangoproject.com/en/3.0/ref/models/instances/#validating-objects

juliasaveliff: HOW FUN!

rackcity/utils/change_planner_utils.py
juliasaveliff: I thought so, just giving u extra work 😘

PR #423, opened 2020-04-13
throw error when deleting a chassis with blades in it

benauriemma:


rackcity/permissions/permissions.py
```diff
@@ -91,7 +91,7 @@ def user_has_asset_permission(user, site=None):
 
 def validate_user_asset_permission_to_add(user, validated_data):
     # TODO: make this check work for blades, which don't have a rack
-    if "rack" not in validated_data:
+    if "rack" not in validated_data or not validated_data["rack"]:
```
benauriemma, line 94: temporary bug fix until julia updates the permissions

src/components/elementView/detailedView/assetView/assetView.tsx
```diff
@@ -270,8 +270,8 @@ export class AssetView extends React.PureComponent<
                 onConfirm={this.handleDecommission}
               >
                 <p>
-                  Are you sure you want to decommission this asset? This action
-                  cannot be undone.
+                  Are you sure you want to decommission this asset? If it's a
+                  chassis, its blades will also be decommissioned.
```
benauriemma, line 274: This snuck in here from another PR

rackcity/views/asset_views.py
```diff
@@ -494,6 +491,16 @@ def asset_delete(request):
             {"failure_message": Status.AUTH_ERROR + str(auth_error)},
             status=HTTPStatus.UNAUTHORIZED,
         )
+    if existing_asset.model.is_blade_chassis():
+        blades = Asset.objects.filter(chassis=existing_asset)
+        if len(blades)>0:
```
clairecfu, line 496: nit:  you could just do `if Asset.objects.filter(chassis=existing_asset).exists():`

PR #424, opened 2020-04-13
Update power endpoints for blades

juliasaveliff:
- Add more robust checks to power endpoints
- Update logs

### DIFF
--- a/rackcity/views/power_views.py
+++ b/rackcity/views/power_views.py
@@ -59,7 +59,7 @@ def pdu_power_status(request, id):
         return JsonResponse(
             {
                 "failure_message": Status.ERROR.value
-                + "Power is only network controllable for assets in datacenter 'RTP1'."
+                + "Power is not network controllable on this rack."
             },
             status=HTTPStatus.BAD_REQUEST,
         )
@@ -112,7 +112,7 @@ def pdu_power_on(request):
     except UserPowerPermissionException as error:
         return JsonResponse(
             {
-                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
+                "failure_message": Status.AUTH_ERROR.value + AuthFailure.POWER.value,
                 "errors": str(error),
             },
             status=HTTPStatus.UNAUTHORIZED,
@@ -166,7 +166,7 @@ def pdu_power_off(request):
     except UserPowerPermissionException as error:
         return JsonResponse(
             {
-                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
+                "failure_message": Status.AUTH_ERROR.value + AuthFailure.POWER.value,
                 "errors": str(error),
             },
             status=HTTPStatus.UNAUTHORIZED,
@@ -217,7 +217,7 @@ def pdu_power_cycle(request):
     except UserPowerPermissionException as error:
         return JsonResponse(
             {
-                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
+                "failure_message": Status.AUTH_ERROR.value + AuthFailure.POWER.value,
                 "errors": str(error),
             },
             status=HTTPStatus.UNAUTHORIZED,
@@ -322,7 +322,7 @@ def remove_unavailable_pdu_ports(
                     raise Exception(
                         "Port "
                         + asset_power[port_num]["port_number"]
-                        + " does not exist on PDU"
+                        + " does not exist on PDU. "
                     )
             else:
                 try:
@@ -331,7 +331,7 @@ def remove_unavailable_pdu_ports(
                     raise Exception(
                         "Port "
                         + asset_power[port_num]["port_number"]
-                        + " does not exist on PDU"
+                        + " does not exist on PDU. "
                     )
 
 
@@ -383,7 +383,7 @@ def chassis_power_status(request):
     except UserPowerPermissionException as error:
         return JsonResponse(
             {
-                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
+                "failure_message": Status.AUTH_ERROR.value + AuthFailure.POWER.value,
                 "errors": str(error),
             },
             status=HTTPStatus.UNAUTHORIZED,
@@ -430,7 +430,7 @@ def chassis_power_on(request):
     except UserPowerPermissionException as error:
         return JsonResponse(
             {
-                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
+                "failure_message": Status.AUTH_ERROR.value + AuthFailure.POWER.value,
                 "errors": str(error),
             },
             status=HTTPStatus.UNAUTHORIZED,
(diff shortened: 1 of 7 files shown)


rackcity/views/power_views.py
```diff
@@ -443,15 +373,22 @@ def toggle_pdu_power(asset, asset_port_number, goal_state):
...
+    except PowerManagementException as error:
         return JsonResponse(
             {"failure_message": Status.ERROR.value + str(error)},
             status=HTTPStatus.BAD_REQUEST,
         )
-    result, exit_status = make_bcman_request(chassis_hostname, str(blade_slot), "")
+    except UserPowerPermissionException as error:
+        return JsonResponse(
+            {
+                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
```
benauriemma, line 386: I was wrong last time but I THINK this should be `Status.AUTH_ERROR.value`

rackcity/views/power_views.py
```diff
@@ -475,21 +412,30 @@ def chassis_power_status(request):
...
+    except PowerManagementException as error:
         return JsonResponse(
             {"failure_message": Status.ERROR.value + str(error)},
             status=HTTPStatus.BAD_REQUEST,
         )
-    result, exit_status = make_bcman_request(chassis_hostname, str(blade_slot), "on")
+    except UserPowerPermissionException as error:
+        return JsonResponse(
+            {
+                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
```
benauriemma, line 433: Ditto

rackcity/views/power_views.py
```diff
@@ -508,15 +455,22 @@ def chassis_power_on(request):
...
+    except PowerManagementException as error:
         return JsonResponse(
             {"failure_message": Status.ERROR.value + str(error)},
             status=HTTPStatus.BAD_REQUEST,
         )
-    result, exit_status = make_bcman_request(chassis_hostname, str(blade_slot), "off")
+    except UserPowerPermissionException as error:
+        return JsonResponse(
+            {
+                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
```
benauriemma, line 468: Ditto

rackcity/views/power_views.py
```diff
@@ -535,16 +490,23 @@ def chassis_power_off(request):
...
+        chassis, blade_slot = get_chassis_power_request_parameters(request)
+    except PowerManagementException as error:
         return JsonResponse(
             {"failure_message": Status.ERROR.value + str(error)},
             status=HTTPStatus.BAD_REQUEST,
         )
+    except UserPowerPermissionException as error:
+        return JsonResponse(
+            {
+                "failure_message": Status.AUTH_ERROR + AuthFailure.POWER.value,
```
benauriemma, line 503: Ditto

rackcity/views/power_views.py
```diff
@@ -596,30 +558,75 @@ def make_bcman_request(chassis, blade, power_command):
...
+            + " does not have power permission and does not own asset with id="
+            + str(data["id"])
         )
-    if (not chassis.model.is_blade_chassis) or (chassis.model.vendor != "BMI"):
-        raise ChassisPowerManagementException(
-            "Power is only network controllable for blade chassis of vendor 'BMI'."
+    if not is_asset_power_controllable_by_bcman(chassis):
+        raise PowerManagementException(
+            "Power is only network controllable for blade chassis of vendor 'BMI' with hostname not in storage."
         )
```
benauriemma, line 588: I've mostly been using the convention of ending Exception messages with a period and a space so they can be chained together, but if you know that these are never followed by anything then I'm fine with it

benauriemma: Goes for all the exceptions you raise in this PR

rackcity/views/power_views.py
```diff
@@ -596,30 +558,75 @@ def make_bcman_request(chassis, blade, power_command):
...
+
+
+def get_pdu_power_request_parameters(request):
+    data = JSONParser().parse(request)
+    if "id" not in data:
+        raise PowerManagementException("Must specify 'id' on pdu power request.")
+    try:
+        asset = Asset.objects.get(id=data["id"])
+    except ObjectDoesNotExist:
+        raise PowerManagementException("Asset" + GenericFailure.DOES_NOT_EXIST.value)
```
benauriemma, line 603: This I think should have a space

benauriemma: After Asset

rackcity/views/power_views.py
```diff
@@ -596,30 +558,75 @@ def make_bcman_request(chassis, blade, power_command):
...
+    if not user_has_power_permission(request.user, asset=asset):
+        raise UserPowerPermissionException(
+            "User "
+            + request.user.username
+            + " does not have power permission and does not own asset with id="
+            + str(data["id"])
+        )
+    if not is_asset_power_controllable_by_pdu(asset):
+        raise PowerManagementException(
+            "Power is only network controllable for assets in datacenter 'RTP1'."
```
benauriemma, line 613: Picky but I think only some rack in RTP1 have networked PDUs if you want to specify that or just change to 
`Power is not network controllable on this rack. `.  I think this message might have been repeated earlier as well

rackcity/views/power_views.py
```diff
@@ -596,30 +558,75 @@ def make_bcman_request(chassis, blade, power_command):
...
+        and asset.hostname is not None
+    )
+
+
+def is_asset_power_controllable_by_pdu(asset):
+    return (
+        asset.model.is_rackmount()
+        and not asset.is_in_offline_storage()
+        and asset.rack.is_network_controlled
+    )
```
benauriemma, line 632: I think these are good candidates for class functions on the model but I'm fine with them here as well

juliasaveliff: ok yeah I TRIED THAT because I thought so too but it wouldn't let me call `self.model.is_blade_chassis()` or any other methods on the model field, it was expecting a `self` parameter to be passed in, which ain't how class methods work. didn't investigate why but maybe it's a model thing?

rackcity/views/power_views.py
juliasaveliff: UR SO RIGHT

rackcity/views/power_views.py
juliasaveliff: fixed

rackcity/views/power_views.py
juliasaveliff: FIXED

rackcity/views/power_views.py
juliasaveliff: fixed

rackcity/views/power_views.py
juliasaveliff: ok word will do that

rackcity/views/power_views.py
juliasaveliff: ACTUALLY i did something crazy and made that enum equal to `DOES_NOT_EXIST = " does not exist."` so you never have to add a space

rackcity/views/power_views.py
juliasaveliff: fair enough. updated this

rackcity/views/power_views.py
benauriemma: CRAZY

rackcity/views/power_views.py
benauriemma: Wow great minds

PR #425, opened 2020-04-13
Update asset permissions for blades

juliasaveliff:
- Move all user permission checks to validation methods only (and call these ASAP in relevant endpoints) 
    - ie remove redundant checks from `validate_location_modification` method
- One validation method for new asset data (to use when adding asset, or applying changes from data in asset modify) 
- One validation method for existing asset (to use when modifying, deleting, decommissioning existing asset)


src/components/elementView/elementTabContainer.tsx
```diff
@@ -173,7 +173,17 @@ class ElementTabContainer extends React.Component<
...
+          className="tab-header do-not-print"
+          id="sites"
+          title="Sites"
+          disabled={
+            !(
+              this.props.permissionState.admin ||
+              this.props.permissionState.asset_management
+            )
+          }
+        />
```
juliasaveliff, line 186: needed to add disabling logic to Site tab header

rackcity/utils/rackcity_utils.py
```diff
@@ -174,53 +173,37 @@ def validate_location_modification(data, existing_asset, user, change_plan=None)
...
     asset_height = existing_asset.model.height
-    if "rack_position" in data:
+    if "rack_position" in data and data["rack_position"]:
         try:
             asset_rack_position = int(data["rack_position"])
         except ValueError:
             raise Exception("Field 'rack_position' must be of type int.")
 
-    if "chassis_slot" in data:
+    if "chassis_slot" in data and data["chassis_slot"]:
```
juliasaveliff, line 182: these values are sent as null from the front end now so casting to `int` was causing uncaught error. also BTW for everyone ValueError catches non-int problems but it doesn't catch None type objects (that throws a TypeError)

rackcity/utils/errors_utils.py
```diff
@@ -16,7 +16,7 @@ class Status(Enum):
     IMPORT_ERROR = "IMPORT ERROR: "
     EXPORT_ERROR = "EXPORT ERROR: "
     CONNECTION = "CONNECTION ERROR: "
-    AUTH_ERROR = "AUTHORIZATION ERROR: "
+    AUTH_ERROR = "PERMISSIONS ERROR: "
```
juliasaveliff, line 19: this looks better

rackcity/utils/asset_utils.py
```diff
@@ -403,11 +403,11 @@ def save_all_connection_data(data, asset, user, change_plan=None):
 def save_all_field_data_live(data, asset):
     asset_id = data["id"]
     for field in data.keys():
-        if field == "model":
+        if field == "model" and data["model"]:
```
juliasaveliff, line 406: same as in rackcity_utils.py

rackcity/permissions/permissions.py
```diff
@@ -75,48 +75,90 @@ def get_permission(permission_name: PermissionName) -> Permission:
...
+    return user.has_perm(
+        PermissionPath.ASSET_WRITE.value
+    ) or site.user_has_site_level_permission(user)
+
+
+def get_id_from_data(data, field, data_is_validated):
+    if data_is_validated:
+        return data[field].id
+    else:
+        return data[field]
```
juliasaveliff, line 88: added this because we call it from asset_add with serializer validated data, which serializes foreign keys to their models so you need to access id. but we also call it from asset_modify with the straight up data that is sent so foreign keys are listed as their id.

rackcity/permissions/permissions.py
```diff
@@ -75,48 +75,90 @@ def get_permission(permission_name: PermissionName) -> Permission:
     return permission
 
 
-def user_has_asset_permission(user, site=None):
-    if user.has_perm(PermissionPath.ASSET_WRITE.value):
-        return True
-    if site:
+def user_has_asset_permission(user, site):
+    return user.has_perm(
```
benauriemma, line 79: Generally opposed to abbreviations in code but I like `user.has_perm` bc it makes me think of a user having a perm like the hair style so I'll let it slide

rackcity/utils/errors_utils.py
benauriemma: Like

rackcity/utils/rackcity_utils.py
benauriemma: Nice catches

rackcity/views/asset_views.py
```diff
@@ -485,19 +494,19 @@ def asset_delete(request):
...
         )
     if existing_asset.model.is_blade_chassis():
         blades = Asset.objects.filter(chassis=existing_asset)
-        if len(blades)>0:
+        if len(blades) > 0:
             return JsonResponse(
                 {
                     "failure_message": Status.DELETE_ERROR.value
-                                       + "Cannot delete a chassis with blades in it. "
+                    + "Cannot delete a chassis with blades in it. "
```
benauriemma, line 509: Looks like someoneeeeee (me) didn't lint their last PR

src/components/elementView/elementTabContainer.tsx
benauriemma: True dat

rackcity/permissions/permissions.py
juliasaveliff: HAHAHA please take this up with Django, sir. I did not write the method.

PR #431, opened 2020-04-15
Brebdab mobile front end + camera

brendancheng20:
- [ ] Set up mobile browser detection
- [ ] Removed non-mobile routes (everything except login and scanner, will implement mobile assetView once the endpoint is set up)
- [ ] Camera integration - Must be over HTTPS
- [ ] Updated navbar to only show mobile-relevant options, so basically home (which redirects to scanner instead of dashboard, will probably change to a different button once similar functionality exists on the desktop version) and logout/login


djreact/settings.py
```diff
@@ -14,10 +14,11 @@
     "rack-city-dev.herokuapp.com",
     "rack-city-staging.herokuapp.com",
     "rack-city-prod.herokuapp.com",
+    "rack-city-mobile.herokuapp.com",
     "127.0.0.1",
     "localhost",
-    "192.168.1.17",
-    "192.168.0.18",
+    # "192.168.1.17",
```
brendancheng20, line 20: commented these out since there seemed to be some weird routing thing going on with heroku if I kept them in

rackcity/views/asset_views.py
```diff
@@ -1001,3 +1001,16 @@ def asset_number(request):
...
+def get_asset_from_barcode(request):
+    """
+    Parse barcode as base64 encoded string for asset number then return
+    corresponding live asset with that number, if it exists
+    """
+    data = JSONParser().parse(request)
+    if "img_string" in data:
+        return JsonResponse({"img_string": data.img_string}, status=HTTPStatus.OK)
+    else:
+        return JsonResponse({"failure_message": "failed"}, status=HTTPStatus.BAD_REQUEST)
```
brendancheng20, line 1016: This is kinda just a dummy method for now. This will be the main backend function that I write though

src/components/mobile/barcodeScanner.tsx
```diff
@@ -29,42 +35,86 @@ export class BarcodeScanner extends React.PureComponent<
...
-
-  render() {
-    const constraints = {
-      height: 720,
-      width: 1280,
-      facingMode: "user",
-      //   facingMode: { exact: "environment" }
+  constraints = {
+    height: 720,
+    width: 1280,
```
brendancheng20, line 40: These are hardcoded in bc I think they basically determine the image quality and not the actual size of the component as it renders? It was weird trying to get the camera to be the right size so I'm not entirely positive how this works but that seems to be it

src/components/mobile/barcodeScanner.tsx
```diff
@@ -29,42 +35,50 @@ export class BarcodeScanner extends React.PureComponent<
...
+            height={this.state.cameraHeight*0.8}
             screenshotFormat={"image/jpeg"}
-            width={1280}
-            videoConstraints={constraints}
-            mirrored={true}
+            width={this.state.cameraWidth}
+            videoConstraints={this.constraints}
+            ref={webcamRef}
           />
           <button onClick={capture}>Capture photo</button>
```
brendancheng20, line 64: forgot to make  a task to make this an AnchorButton. Will do next PR

PR #435, opened 2020-04-15
prevent deletion of hostname on a chassis with blades in it

benauriemma:


benauriemma:
I didn't add logic to prevent deletion of a hostname on assets with network connections because that's not new and I don't think they'll add it to the test plan

PR #440, opened 2020-04-16
SQUASHING some change plan bugs 

clairecfu:
image


rackcity/utils/change_planner_utils.py
```diff
@@ -85,15 +87,14 @@ def detect_conflicts_cp(sender, **kwargs):
             & Q(change_plan__execution_time=None)
         ):
             try:
-                # TODO: add check for blades
                 validate_asset_location_in_rack(
                     asset.rack_id,
                     asset_cp.rack_position,
                     asset_cp.model.height,
                     asset_id=asset_cp.id,
+                    related_asset_id=asset_cp.related_asset_id,
```
clairecfu, line 95: the SOURCE OF THE rack location bug

rackcity/utils/change_planner_utils.py
```diff
@@ -85,15 +87,14 @@ def detect_conflicts_cp(sender, **kwargs):
...
                 )
-            except LocationException:
-                asset_cp.asset_conflict_location = asset
```
clairecfu, line 96: this line is redundant because the next line does the updating

rackcity/utils/change_planner_utils.py
```diff
@@ -70,11 +71,12 @@ def detect_conflicts_cp(sender, **kwargs):
 
     # asset hostname conflicts with hostnames on an active assetCP
     asset.hostname_conflict.clear()
-    AssetCP.objects.filter(
-        Q(hostname=asset.hostname)
-        & ~Q(related_asset=asset.id)
-        & Q(change_plan__execution_time=None)
-    ).update(asset_conflict_hostname=asset)
+    if asset.hostname:
```
clairecfu, line 74: hostname bug

rackcity/utils/change_planner_utils.py
```diff
@@ -147,16 +151,24 @@ def get_change_plan(change_plan_id):
...
             racked_assets = racked_assets.filter(~Q(id=asset_cp.related_asset.id))
+    ## don't return decommissioned asset_cps:
+    racked_assets_cp = racked_assets_cp.filter(is_decommissioned=False)
     return racked_assets, racked_assets_cp
 
 
 def get_offline_storage_assets_for_cp(change_plan):
     stored_assets = Asset.objects.filter(offline_storage_site__isnull=False)
-    stored_assets_cp = AssetCP.objects.filter(change_plan=change_plan, offline_storage_site__isnull=False)
+    stored_assets_cp = AssetCP.objects.filter(
```
clairecfu, line 167: ignore decomissioned assets

rackcity/utils/change_planner_utils.py
```diff
@@ -230,8 +244,11 @@ def get_many_assets_response_for_cp(request, change_plan, decommissioned=False,
     )
     if filter_failure_response:
         return filter_failure_response
+    if decommissioned:
+        asset_serializer = GetDecommissionedAssetSerializer(assets, many=True,)
```
clairecfu, line 248: decommissioned assets need to use diff serializer

rackcity/utils/rackcity_utils.py
```diff
@@ -85,7 +85,7 @@ def validate_asset_location_in_rack(
                     if asset_in_rack.asset_number:
                         raise LocationException(
                             "Asset location conflicts with another asset: '"
-                            + str(asset_in_rack.asset_number)
+                            + str(asset_in_rack.id)
```
clairecfu, line 88: fuk i need to change this back.

rackcity/utils/change_planner_utils.py
juliasaveliff: i think u wanna do the same thing here that u did above in `get_racked_assets_for_cp` and filter out decommissioned AssetCP's AFTER filtering out all change-plannified stored Asset's. cuz if u have stored Asset live, and u have decommissioned it in a cp so its a stored + decommissioned AssetCP, then u DON'T want to return that Asset in the query so u need the related AssetCP here. but u DONT want to return the AssetCP either so u have to filter it out after

rackcity/utils/change_planner_utils.py
juliasaveliff: so true.

rackcity/utils/change_planner_utils.py
clairecfu: good point

PR #441, opened 2020-04-16
2 change plan bug fixies

juliasaveliff:


rackcity/utils/change_planner_utils.py
```diff
@@ -362,12 +362,22 @@ def get_cp_modification_conflicts(asset_cp):
     )
     conflicting_asset_message_2 = " now conflicts with a live asset. "
     if asset_cp.is_conflict and asset_cp.related_asset is None:
-        conflicts.append(
-            {
-                "conflict_message": "This asset has been deleted live. "
-                + nonresolvable_message
-            }
-        )
+        if asset_cp.related_decommissioned_asset:
```
juliasaveliff, line 365: if a asset cp was decommissioned live it used to think it was deleted AND decommissioned. fixed

rackcity/views/change_plan_views.py
```diff
@@ -307,7 +307,11 @@ def change_plan_page_count(request):
     Return total number of pages according to page size, which must be
     specified as query parameter.
     """
-    return get_page_count_response(ChangePlan, request.query_params)
+    user = request.user
+    user_change_plans = ChangePlan.objects.filter(owner=user)
+    return get_page_count_response(
```
juliasaveliff, line 312: page count was not being filtered by change plan owner

clairecfu: i think we wrote the EXACT same code to fix this bug lmao
so probably no conflicts

PR #442, opened 2020-04-16
Chassis power inline + power updates

juliasaveliff:
- [x] Inline power for blades in BMI chassis
- [x] Refactor powerView so power status endpoint only called if asset is connected to a PDU that is network controlled

Need to get loading spinners working better but going to do that in another PR


src/components/elementView/powerView/bladePowerView.tsx
```diff
@@ -273,7 +288,7 @@ export class BladePowerView extends React.PureComponent<
         )}
         <Alert
           className={Classes.DARK}
-          confirmButtonText="Okay"
+          confirmButtonText="OK"
```
clairecfu, line 291: small changes

PR #446, opened 2020-04-16
Scanner front end/backend integration

brendancheng20:
- [ ] Scanner front end makes POST request with image string on button click
- [ ] Backend route implemented with error handling

I think this is a fairly small PR in terms of volume of code, but the meat of the camera integration and front/back end integration is here.


rackcity/views/asset_views.py
```diff
@@ -1026,9 +1029,52 @@ def get_asset_from_barcode(request):
...
+    except Asset.DoesNotExist:
+        return JsonResponse(
+            {
+                "failure_message": Status.ERROR.value
+                + "Asset"
+                + GenericFailure.DOES_NOT_EXIST.value,
+                "errors": "No existing asset with asset_number=" + str(barcode_data),
+            },
+            status=HTTPStatus.BAD_REQUEST,
+        )
```
juliasaveliff, line 1075: what do you think about having a more general failure message to show to the user? I feel like the differences don't matter to the user, they just need to know that there's no hyposoft asset associated with the barcode they sent - so maybe something like `"There is no existing asset associated with this barcode"`?

rackcity/views/asset_views.py
```diff
@@ -1026,9 +1029,49 @@ def get_asset_from_barcode(request):
...
+        for barcode in barcodes:
+            barcode_data = barcode.data.decode("utf-8")
+    except Exception as error:
+        return JsonResponse(
+            {"failure_message": "ERROR: Could not decode barcode value"},
+            status=HTTPStatus.BAD_REQUEST,
+        )
+    if not barcode_data.isdigit():
+        return JsonResponse(
+            {"failure_message": "ERROR: Barcode contains non-numeric characters"},
```
juliasaveliff, line 1061: just to clarify i meant returning consistent failure messages here!

rackcity/views/asset_views.py
```diff
@@ -1026,9 +1029,49 @@ def get_asset_from_barcode(request):
...
+            {"warning_message": "WARNING: No barcode detected"},
+            status=HTTPStatus.BAD_REQUEST,
         )
+    barcode_data = ""
+    try:
+        for barcode in barcodes:
+            barcode_data = barcode.data.decode("utf-8")
+    except Exception as error:
+        return JsonResponse(
+            {"failure_message": "ERROR: Could not decode barcode value"},
```
juliasaveliff, line 1056: and maybe here? depending on when this gets code path hit?

PR #447, opened 2020-04-16
Switching between sites + MODIFY ASSET BUG

juliasaveliff:


rackcity/migrations/0052_merge_20200416_1547.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.0.3 on 2020-04-16 19:47
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
juliasaveliff, line 6: nobody ever migrated this poor bby

rackcity/models/asset.py
```diff
@@ -137,10 +137,6 @@ def validate_location_type(
                 "Rackmount assets must not have a chassis or chassis slot. "
             )
     else:
-        if offline_storage_site and (chassis or chassis_slot):
```
juliasaveliff, line 140: per discussions with Benjamin, blades in offline storage will stay in their chassis

rackcity/utils/rackcity_utils.py
```diff
@@ -58,6 +58,8 @@ def validate_asset_location_in_rack(
     change_plan=None,
     related_asset_id=None,
 ):
+    if asset_rack_position is None:
+        return
     new_asset_location_range = [asset_rack_position + i for i in range(asset_height)]
```
juliasaveliff, line 63: THIS LIST COMPREHENSION WAS THROWING AN EXCEPTION DEEP IN THE DEPTHS OF ASSET SAVING

rackcity/models/asset.py
benauriemma: I think they can be either in or out of their chassis?

juliasaveliff: yeah that's wha ti meant

PR #448, opened 2020-04-16
rack elevation diagrams

clairecfu:
support chassis + blades

### DIFF
--- a/src/components/elementView/detailedView/rackView/rackView.scss
+++ b/src/components/elementView/detailedView/rackView/rackView.scss
@@ -14,7 +14,7 @@ $row-height: 6px;
 .rack {
   border-style: solid;
   border-width: 6pt;
-  border-color: #202b33;
+  border-color: #182026;
   display: inline-block;
   margin: left 10px;
 }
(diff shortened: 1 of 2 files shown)


benauriemma:
- When there's one blade is says "1 blades", wanna make that singular?
- If the hostname or model are long, you have to scroll horizontally to see that there are blades.   What would you think about changing the order to make it "host-name | 2 blades | Model Number" so you can always see how many blades there are? You'd have to scroll for model number but I think that's least important
- What would you think about making the chassis' pop a little more? Like bolding or underlining "1 blade" or the whole thing?

clairecfu:
- When there's one blade is says "1 blades", wanna make that singular?
 yes 


- If the hostname or model are long, you have to scroll horizontally to see that there are blades. What would you think about changing the order to make it "host-name | 2 blades | Model Number" so you can always see how many blades there are? You'd have to scroll for model number but I think that's least important
yes
- What would you think about making the chassis' pop a little more? Like bolding or underlining "1 blade" or the whole thing?
i like this idea but i think its low priority bc its slightly more complicated bc this could mess up how it looks when u print /are on light colored backgrounds.

PR #450, opened 2020-04-16
Backend updates for change plan detail

juliasaveliff:
- [x] Do not return non-modified assets in modifications
- [x] Update modification summaries to support all offline / blade / racked assets
- [x] Do not double count decommissioned assets as modified

### DIFF
--- a/rackcity/utils/change_planner_utils.py
+++ b/rackcity/utils/change_planner_utils.py
@@ -359,7 +359,12 @@ def get_changes_on_asset(asset, asset_cp):
     fields = [field.name for field in Asset._meta.fields]
     changes = []
     for field in fields:
-        if getattr(asset, field) != getattr(asset_cp, field):
+        if field == "chassis":
+            chassis_live = asset.chassis
+            chassis_cp = asset_cp.chassis
+            if chassis_live and (chassis_live != chassis_cp.related_asset):
+                changes.append("chassis")
+        elif getattr(asset, field) != getattr(asset_cp, field):
             changes.append(field)
     if network_connections_have_changed(asset, asset_cp):
         changes.append("network_connections")
--- a/rackcity/utils/change_planner_utils.py
+++ b/rackcity/utils/change_planner_utils.py
@@ -468,37 +468,56 @@ def get_cp_modification_conflicts(asset_cp):
         return conflicts
 
 
-def get_location_detail(asset,):
-    if asset.model.is_rackmount():
-        if asset.rack:
-            return (
-                " at rack "
-                + asset.rack.datacenter.abbreviation
-                + " "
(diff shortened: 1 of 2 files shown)


rackcity/utils/change_planner_utils.py
```diff
@@ -359,7 +359,12 @@ def get_changes_on_asset(asset, asset_cp):
     fields = [field.name for field in Asset._meta.fields]
     changes = []
     for field in fields:
-        if getattr(asset, field) != getattr(asset_cp, field):
+        if field == "chassis":
+            chassis_live = asset.chassis
+            chassis_cp = asset_cp.chassis
+            if chassis_live and (chassis_live != chassis_cp.related_asset):
+                changes.append("chassis")
+        elif getattr(asset, field) != getattr(asset_cp, field):
```
juliasaveliff, line 367: chassis should be the only field we have to add extra logic to check if they're different on asset + assetcp because asset's chassis is an Asset and assetcp's chassis is an AssetCP

rackcity/utils/change_planner_utils.py
```diff
@@ -468,37 +473,56 @@ def get_cp_modification_conflicts(asset_cp):
...
     modifications = []
     for asset_cp in assets_cp:
-        asset_cp_data = RecursiveAssetCPSerializer(asset_cp).data
         related_asset = asset_cp.related_asset
         if related_asset:
-            asset_data = RecursiveAssetSerializer(related_asset).data
             changes = get_changes_on_asset(related_asset, asset_cp)
+            if changes is None and not asset_cp.is_decommissioned:
+                continue
+            asset_data = RecursiveAssetSerializer(related_asset).data
```
juliasaveliff, line 521: reused the existing `get_changes_on_asset` method 😅if no fields have changed AND the assetcp hasn't been decommissioned, then don't show it in the modifications

rackcity/views/change_plan_views.py
```diff
@@ -416,7 +416,7 @@ def change_plan_execute(request, id):
             log_action(
                 request.user, updated_asset, Action.CREATE, change_plan=change_plan,
             )
-        else:
+        elif not asset_cp.is_decommissioned:
```
juliasaveliff, line 419: only count modifications on non-decommissioned assets (so if it's modified then decommissioned in a change plan, we'll consider that a decommission, even though the modifications are still applied)

rackcity/utils/change_planner_utils.py
clairecfu: lol forgot we already had this helper method... also i think we should also check if an asset is being moved between offline sites?

clairecfu: wait jk nvm

PR #451, opened 2020-04-17
Report updates

juliasaveliff:
- Backend support for blades
- Front end redesign


rackcity/views/report_views.py
```diff
@@ -45,6 +46,8 @@ def compute_rack_report(datacenter_id=None):
     owner_allocation = {}
     num_full_rack_slots = 0
     for asset in assets:
+        if asset.model.is_blade_asset():
+            continue
```
benauriemma, line 50: I think this is unnecessary because you filtered for only assets with a rack but it's fine

rackcity/views/report_views.py
```diff
@@ -60,34 +63,30 @@ def compute_rack_report(datacenter_id=None):
...
-        for owner in owner_allocation
-    ]
+    rack_usage_data = {
+        "free_rackspace_percent": 1 - (num_full_rack_slots / total_num_rack_slots),
+        "vendor_allocation": [
+            {
+                "vendor": vendor,
+                "allocation_percent": vendor_allocation[vendor] / num_full_rack_slots,
+            }
+            for vendor in vendor_allocation
```
benauriemma, line 73: SO pythonic of you

rackcity/views/report_views.py
juliasaveliff: oh right hehe oh well

rackcity/views/report_views.py
juliasaveliff: thank you, pycharm encouraged me to do so.

PR #460, opened 2020-04-17
Mobile Asset View

brendancheng20:
- [ ] Implemented mobileAssetView.tsx, which has a lot of similar components/rendering methods to assetView but with some reformatting/refactoring to account for the mobile app structure
- [ ] Entire mobile app (aside from login) is now hosted entirely on one page with no difference in routing for scanner vs. assetView. The reason for this is because the scanner component uses React hooks for capturing the image. For some reason hooks make using redux difficult between routes, so I decided to just keep everything on the one route so I could more easily access things like `token`.
- [ ] On image capture no longer shows alerts, now shows either error toast or success dialog asking if you want to view the asset whose barcode was detected. Clicking yes will render the MobileAssetView.
- [ ] Integrated network connections table and non-blade power connections table with MobileAssetView.
- [ ] Updated navbar scanner button so that user is still redirected to scanner when it is clicked despite everything being on same route.

Still TODO: integrate blade powerview


src/components/mobile/barcodeScanner.tsx
```diff
@@ -53,10 +89,16 @@ export class BarcodeScanner extends React.PureComponent<
           getHeaders(token)
         )
         .then((res: any) => {
-          alert(JSON.stringify(res));
+          const asset_data = res.data.asset_data;
+          this.setState({
+            token: token,
```
brendancheng20, line 94: I addressed this in my overall PR comments but I want to clarify a little more here. For some reason, since this component uses the `useRef()` hook, `MapStatetoProps()` and `connect()` redux functions do not seem to be linking `token` to `this.props.token`. Because of this, I had to use the redux `useSelector()` hook, which is basically the functional programming equivalent of `MapStateToProps()`. I am able to store it in the `BarcodeScanner` state but can't find a way to save it in props. This seems to propagate to other components rendered from this page even if they do not have any  hooks, so I just decided to use `this.state.token` for all of our token needs and pass it in to components on the same page as a prop if needed.

src/components/mobile/barcodeScanner.tsx
```diff
@@ -84,6 +126,29 @@ export class BarcodeScanner extends React.PureComponent<
...
+  private renderAssetView(asset: AssetObject | null) {
+    if (asset) {
+      return (
+        <MobileAssetView
+          {...this.props}
+          token={this.state.token}
+          asset={asset}
+          data_override={() => {
+            const { cpu, display_color, storage, memory_gb } = asset;
+            return { cpu, display_color, storage, memory_gb };
```
brendancheng20, line 138: @clairecfu I wasn't sure how `data_override` works in AssetView but I copied it over to MobileAssetView bc I figured it would still be relevant. Is this correct? Rather than having something like `this.props.asset`, I just pass in the detected asset, which is in `this.state.asset`.

src/components/mobile/mobileAssetView.tsx
```diff
@@ -0,0 +1,302 @@
+import { Callout, Classes } from "@blueprintjs/core";
```
brendancheng20, line 1: There might be some CP/decommission/etc. code in here that is reused from AssetView, but that's mostly because i don't fully understand the rendering logic involving CP/decommission/any other new features we've implemented and didn't want to break anything. It seems to work fine though.

src/components/mobile/mobileAssetView.tsx
```diff
@@ -0,0 +1,302 @@
...
+          //     token={this.props.token}
+          //   asset={this.props.asset}
+          //   shouldUpdate={this.state.powerShouldUpdate}
+          //   updated={() => {
+          //     this.setState({ powerShouldUpdate: false });
+          //   }}
+          //   assetIsDecommissioned={
+          //     this.props.asset.decommissioning_user !== undefined
+          //   }
+          // />
```
brendancheng20, line 181: Currently commented out because I have not updated BladePowerView to fit mobile requirements, but will do in next PR

src/components/mobile/mobileAssetView.tsx
```diff
@@ -0,0 +1,302 @@
...
+                                //     ? undefined
+                                //     : (e: any) => {
+                                //         const id = this.getAssetIdFromHostname(
+                                //           connection!.destination_hostname!
+                                //         );
+                                //         if (id) {
+                                //           this.redirectToAsset(id);
+                                //         }
+                                //       }
+                                // }
```
brendancheng20, line 267: This is worth discussing. Do we want to be able to click to other assets in mobile? My gut tells me it probably doesn't matter one way or another, since the sense I got from the requirements was just that the mobileAssetView only needs to display "pertinent information" about the asset and does not specify any need for this functionality.

src/components/mobile/mobileAssetView.tsx
```diff
@@ -0,0 +1,302 @@
...
+                </tbody>
+              </table>
+
+              <NetworkGraph
+                networkGraph={this.props.asset.network_graph}
+                onClickNode={() => {}}
+                isDecommissioned={
+                  this.props.asset.decommissioning_user !== undefined
+                }
+              />
```
brendancheng20, line 290: This does not render on mobile. Maybe it's a platform issue where the Network Graph component is just not compatible, idk. Do you think we should show this or remove it? The requirements again do not specify other than to say that we should show "pertinent information about the asset"

src/components/mobile/mobileAssetView.tsx
juliasaveliff: if that's what the requirements say then I would lean towards no, I think this asset view can be simpler than the other one. It's probably worth clarifying this on piazza though?

src/components/mobile/mobileAssetView.tsx
juliasaveliff: good piazza Q 😁

src/components/mobile/mobileAssetView.tsx
juliasaveliff: the user will have no way to enter a change plan from mobile so you should be safe to remove all that logic

src/components/elementView/powerView/powerView.tsx
```diff
@@ -295,7 +296,8 @@ export class PowerView extends React.PureComponent<
         {this.props.asset &&
         this.state.powerStatus &&
         this.state.statusLoaded &&
-        this.shouldShowPowerButtons() ? (
+        this.shouldShowPowerButtons() &&
+        !this.props.isMobile ? (
```
juliasaveliff, line 300: so your plan is to show power status, but not allow for control of power from mobile, right?

rackcity/views/asset_views.py
```diff
@@ -1086,7 +1086,7 @@ def get_asset_from_barcode(request):
     barcodes = decode(opencv_img)
     if len(barcodes) < 1:
         return JsonResponse(
-            {"warning_message": "WARNING: No barcode detected"},
+            {"failure_message": "WARNING: No barcode detected"},
```
juliasaveliff, line 1089: picky but we probably don't need to show warning if it's not a warning :D

src/components/elementView/powerView/powerView.tsx
brendancheng20: Right!

rackcity/views/asset_views.py
brendancheng20: Will fix in next push

juliasaveliff: think u missed this one!

PR #464, opened 2020-04-18
Default sorting on change plan

juliasaveliff:
now change plan assets will ALWAYS be returned first in tables (assets + stored + decommissioned), and the sorting will be applied separately to the two groups of assets


benauriemma:
I would have expected that change plan coming first was just a default, but once they got sorted they would be mixed in - is that not the case? I don't hate it if they always come first as long as we can justify it to Bletsch

juliasaveliff:
@benauriemma oh. we are on the same page in a different book. that is not the case. change planned assets always appear at the top. the query is more optimized now at least...  aha ah..

PR #469, opened 2020-04-18
beautify network graph

clairecfu:


benauriemma:
Going to approve but a couple questions:
- It's really big now - is that intentional?
- Every time I refresh the nodes show up in different places, seems like it's random. Idk if you can but it would be nice if that was more deterministic
- Have you checked that this still works on dev?

PR #470, opened 2020-04-18
Fix model view for custom assets 

clairecfu:
image


src/components/elementView/detailedView/assetView/assetView.tsx
```diff
@@ -382,15 +387,7 @@ export class AssetView extends React.PureComponent<
...
-              data_override={() => {
-                const {
-                  cpu,
-                  display_color,
-                  storage,
-                  memory_gb,
-                } = this.state.asset;
-                return { cpu, display_color, storage, memory_gb };
-              }}
+              data_override={this.getDataOverride()}
```
brendancheng20, line 390: can you clarify how data_override works in PropertiesView? I want to make sure I'm coding it right for mobile, right now I basically just copied what you have here I think

PR #472, opened 2020-04-19
Chassis diagram (frontend) + Bulk tabs/css

clairecfu:
image
- [x] also some navbar css changes

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -154,9 +154,9 @@ class App extends React.Component<AppProps> {
                 component={ModelView}
               />
               <this.PrivateRoute
-              path={ROUTES.BULK_IMPORT}
-              component={BulkImport}
-            />
+                path={ROUTES.BULK_IMPORT}
+                component={BulkImport}
+              />
               <this.PrivateRoute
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -161,23 +165,21 @@ class App extends React.Component<AppProps> {
                 path={ROUTES.CHANGE_PLAN + "/:id"}
                 component={CPDetailView}
               />
+
             </div>
+                                      <this.PrivateRoute
```
brendancheng20, line 170: probs a little picky but is this a linting thing we can fix?

src/components/elementView/detailedView/assetView/assetView.scss
```diff
@@ -60,7 +62,7 @@ margin:auto;
 
   max-width:30pt;
   min-width:30pt;
-  border: 1px solid #BFCCD6 !important;
+  border: 1px solid #BFCCD6;
```
brendancheng20, line 65: is removing the `!important` tag to allow for highlighting of chassis diagrams?

clairecfu: yes!

PR #473, opened 2020-04-19
Barcode printing white background

brendancheng20:


rackcity/migrations/0054_merge_20200417_2324.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('rackcity', '0053_auto_20200416_2333'),
+        ('rackcity', '0052_merge_20200416_1547'),
+    ]
+
+    operations = [
+    ]
```
brendancheng20, line 14: These were just migrations I needed to make to be able to deploy on heroku.

src/components/elementView/detailedView/assetView/barcodeView.tsx
```diff
@@ -25,6 +25,7 @@ class BarcodeView extends React.PureComponent<
   }
 
   public render() {
+    document.body.style.background = "white";
```
brendancheng20, line 28: On render, set the background color to white

src/components/elementView/detailedView/assetView/barcodeView.tsx
```diff
@@ -38,6 +39,10 @@ class BarcodeView extends React.PureComponent<
     );
   }
 
+  public componentWillUnmount(): void {
+    document.body.style.background = "#293742";
+  }
+
```
brendancheng20, line 45: Since navbar renders on printing page and the user can therefore technically navigate back to the dashboard from here, I return the background color from white to it's original state before officially navigating to another page.

clairecfu: gotcha -> it might be worth just removing the navbar from this page all together but I can track that since if we do that we should probably also do it on rack print.

PR #476, opened 2020-04-19
bulk changes for files 3, 13, 14, plus an error message and a front end title

benauriemma:
…nd title

### DIFF
--- a/rackcity/views/asset_views.py
+++ b/rackcity/views/asset_views.py
@@ -1,6 +1,6 @@
 from base64 import b64decode
 import csv
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.http import JsonResponse
 from http import HTTPStatus
 from io import StringIO, BytesIO
@@ -23,6 +23,7 @@ from rackcity.models import (
     ITModel,
     Rack,
     Site,
+    validate_location_type,
 )
 from rackcity.models.asset import (
     get_assets_for_cp,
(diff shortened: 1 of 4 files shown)


rackcity/views/asset_views.py
```diff
@@ -795,6 +796,24 @@ def asset_bulk_upload(request):
...
+            )
+        except ValidationError as error:
+            return JsonResponse(
+                {
+                    "failure_message": Status.IMPORT_ERROR.value
+                    + BulkFailure.ASSET_INVALID.value
+                    + str(error.message)
+                },
+                status=HTTPStatus.BAD_REQUEST,
+            )
```
benauriemma, line 816: This gets called on save but that is too! late!

rackcity/views/it_model_views.py
```diff
@@ -397,6 +401,35 @@ def model_bulk_upload(request):
...
+            )
+        except ValidationError as error:
+            return JsonResponse(
+                {
+                    "failure_message": Status.IMPORT_ERROR.value
+                    + BulkFailure.MODEL_INVALID.value
+                    + str(error.message)
+                },
+                status=HTTPStatus.BAD_REQUEST,
+            )
```
benauriemma, line 432: Same with these

src/components/import/import.tsx
```diff
@@ -274,7 +274,7 @@ export class BulkImport extends React.PureComponent<
             className={Classes.DARK + " modify-table"}
             usePortal={true}
             isCloseButtonShown={true}
-            title={"Network Connections Modifications Menu"}
+            title={"Network Connections Modifications"}
```
juliasaveliff, line 277: thanks :D

PR #477, opened 2020-04-19
Custom assets on mobile! 

clairecfu:
same bug fix as for desktop


src/components/mobile/mobileAssetView.tsx
```diff
@@ -225,6 +225,11 @@ export class MobileAssetView extends React.PureComponent<
               Object.keys(ModelFieldsTable),
               this.props.asset.model
             )}
+                    <p className="custom-message">
```
brendancheng20, line 228: Is this a linting thing? Everything else looks good otherwise

PR #480, opened 2020-04-19
User improvements

juliasaveliff:
- [x] Checkbox align on user permissions form 
- [x] Tooltip for why permissions button is disabled
- [x] Disable permissions button in change plan

### DIFF
--- a/src/components/elementView/elementTable.tsx
+++ b/src/components/elementView/elementTable.tsx
@@ -12,6 +12,7 @@ import {
   Position,
   Spinner,
   Toaster,
+  Tooltip,
 } from "@blueprintjs/core";
 import "@blueprintjs/core/lib/css/blueprint.css";
 import { IconNames } from "@blueprintjs/icons";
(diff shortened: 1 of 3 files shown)


src/components/elementView/elementTable.tsx
```diff
@@ -1005,20 +1006,34 @@ class ElementTable extends React.Component<
...
-          });
-        }}
-        disabled={item.username === this.state.username}
-      />
+      <Tooltip
+        content={
+          item.username === this.state.username
+            ? "Cannot edit your own user permissions"
+            : "Cannot edit within change plan"
+        }
```
clairecfu, line 1014: if youre admin you can revoke admin permissions from yourself?

juliasaveliff: are you saying you should be able to do this? I don't think you should be able to. and I think I have implemented it accordingly (¿?)

clairecfu: wait ignore thsi comment not important lol

PR #493, opened 2020-04-19
BCMAN woo

juliasaveliff:


rackcity/utils/bcman.expect
```diff
@@ -1,7 +1,7 @@
 #!/app/.apt/usr/bin/expect
 
 log_user 0
-set timeout 10
+set timeout 20
```
juliasaveliff, line 4: 10 seconds was sometimes too short but working other times, so i increased it

rackcity/views/power_views.py
```diff
@@ -555,8 +557,6 @@ def make_bcman_request(chassis, blade, power_command):
         result = fp.read().splitlines()[0]
         fp.close()
         os.remove("temp.txt")
-    else:
-        result = "temp.txt file does not exist"
```
juliasaveliff, line 559: this is something i had pushed to dev directly (muahaha) while testing so i'm removing it now

rackcity/views/power_views.py
```diff
@@ -389,13 +389,15 @@ def chassis_power_status(request):
...
         return JsonResponse(
             {
                 "failure_message": Status.CONNECTION.value
                 + "Unable to contact network controlled blade chassis power management.",
-                "errors": "Request to bcman exited with non-zero status: "
-                + str(exit_status),
+                "errors": "Request to bcman exited with status: "
+                + str(exit_status)
+                + ", result: "
+                + str(result),
```
juliasaveliff, line 400: more details if KNOCK ON WOOD i need to debug on heroku in the future

rackcity/views/power_views.py
```diff
@@ -389,13 +389,15 @@ def chassis_power_status(request):
             status=HTTPStatus.UNAUTHORIZED,
         )
     result, exit_status = make_bcman_request(chassis.hostname, str(blade_slot), "")
-    if exit_status != 0:
+    if (exit_status != 0) or (result is None):
```
juliasaveliff, line 392: if the request times out, it won't have written the result yet

PR #496, opened 2020-04-19
Change Plan decommissioning 

clairecfu:
image
- [x] also fixed bug where power ports and network connections dont get copied over to AssetCP on decommission. We had solved this for asset modify, so used same logic as that to fix that bug on decommission

### DIFF
--- a/rackcity/views/change_plan_views.py
+++ b/rackcity/views/change_plan_views.py
@@ -126,7 +126,7 @@ def change_plan_remove_asset(request, id):
     asset_cp = data["asset_cp"]
 
     try:
-        asset_cp_model = AssetCP.objects.get(id=asset_cp)
+        asset_cp_object = AssetCP.objects.get(id=asset_cp)
     except ObjectDoesNotExist:
         return JsonResponse(
             {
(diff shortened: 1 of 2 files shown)


rackcity/views/change_plan_views.py
```diff
@@ -138,7 +138,25 @@ def change_plan_remove_asset(request, id):
...
+                asset_cp_object.delete()
+            else:
+                for field in asset_cp.related_asset._meta.fields:
+                    if (
+                            field.name != "id"
+                            and field.name != "assetid_ptr"
+                            and field.name != "chassis"
+                    ):
+                        setattr(asset_cp, field.name, getattr(asset_cp.related_asset, field.name))
+
```
clairecfu, line 157: if a chassis has blades with changes on the change plan -> dont actually remove it from AssetCP, just make it the same as live .

rackcity/views/change_plan_views.py
```diff
@@ -423,7 +441,7 @@ def change_plan_execute(request, id):
         updated_asset_mappings[asset_cp] = updated_asset
         differs_from_live = True
 
-        if created:
+        if created and not asset_cp.is_decommissioned:
```
clairecfu, line 444: counting bug

rackcity/views/change_plan_views.py
juliasaveliff: this is super fun

rackcity/views/change_plan_views.py
juliasaveliff: it be like. that

PR #505, opened 2020-04-19
Fixed PDU bug if connectivity is lost after loading powerview component

brendancheng20:
No longer getting error where both power success and power failure toasts show up when controlling power if PDU Networx Pro goes down after loading PowerView component

### DIFF
--- a/rackcity/views/power_views.py
+++ b/rackcity/views/power_views.py
@@ -140,7 +140,16 @@ def pdu_power_on(request):
             html.text, power_connections[connection]["port_number"]
         )[0]
         if power_status != "ON":
-            toggle_pdu_power(asset, connection, "on")
+            try:
+                toggle_pdu_power(asset, connection, "on")
+            except ConnectionError:
+                return JsonResponse(
+                    {
+                        "failure_message": Status.CONNECTION.value
+                        + PowerFailure.CONNECTION.value
+                    },
+                    status=HTTPStatus.REQUEST_TIMEOUT,
+                )
     log_power_action(
         request.user, PowerAction.ON, asset,
     )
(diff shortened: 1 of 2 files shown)


rackcity/views/power_views.py
```diff
@@ -140,7 +140,16 @@ def pdu_power_on(request):
             html.text, power_connections[connection]["port_number"]
         )[0]
         if power_status != "ON":
-            toggle_pdu_power(asset, connection, "on")
+            try:
```
brendancheng20, line 143: The issue was that the actual power switching occurs in this `toggle_pdu_power` method, but rather than propagating the `ConnectionError` exception, I was just returning it from that method, so the actual endpoint method never throws an error when attempting to toggle power. I changed the `toggle_pdu_power` call for power switching/cycling methods to be in a try-except block and also changed the method itself to raise an error rather than returning it

src/components/elementView/powerView/powerView.tsx
```diff
@@ -170,6 +170,9 @@ export class PowerView extends React.PureComponent<
           err.response.data.failure_message
         ) {
           this.addErrorToast(err.response.data.failure_message);
+          this.setState({
```
brendancheng20, line 173: Setting the statusLoaded state so that the spinners disappear even if there is a connection error

src/components/elementView/powerView/powerView.tsx
```diff
@@ -283,7 +292,7 @@ export class PowerView extends React.PureComponent<
   private renderPowerTable(loading: boolean) {
     return (
       <div className="power-table">
-        <div className = "network-connections">
+        <div className="network-connections">
```
brendancheng20, line 295: Linting I think

PR #508, opened 2020-04-19
Clear fields on asset form as necessary

clairecfu:


src/forms/assetForm.tsx
```diff
@@ -819,14 +819,15 @@ class AssetForm extends React.Component<AssetFormProps, AssetFormState> {
...
     const newValues = updateObject(this.state.values, {
       rack: undefined,
-      rack_position: null,
       chassis: undefined,
-      chassis_slot: null,
       power_connections: clearedPowerConnections,
       network_connections: clearedNetworkConnections,
     });
+    delete newValues.rack_position;
+    delete newValues.chassis_slot;
```
clairecfu, line 831: for string values, if you set it to null or undefined the form just ignores it and keeps the old value

juliasaveliff: OOOOOO. thx

